PR #2, opened 2017-02-07
Add basic XML tag generation

cfournie:
This PR adds basic XML tag generation for Python 2 and 3 and a basic builder-interface that could be used to submit a simple workflow.

This PR leaves out:
  - Documentation and a clear delineation of the public vs private APIs
  - The actual Oozie API calls

https://github.com/Shopify/pyoozie/issues/3 documents non-implementation details that need to be fixed before releasing.

This PR should serve as a basis for adding Oozie API calls in a future PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,5 +24,6 @@ autopep8:
 lint:
 		@echo 'Linting...'
 		@pylint --rcfile=pylintrc pyoozie tests
+		@flake8
 
 autolint: autopep8 lint
--- a/pyoozie/__init__.py
+++ b/pyoozie/__init__.py
@@ -2,9 +2,9 @@
 # Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 from pyoozie.coordinator import Coordinator, ExecutionOrder
-from pyoozie.tags import validate, Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, \
+from pyoozie.tags import Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, \
     Email, IdentifierTooLongError
-from pyoozie.builder import workflow, coordinator, _coordinator_submission_xml, _workflow_submission_xml
+from pyoozie.builder import workflow, coordinator
 
 __version__ = '0.0.0'
 
@@ -13,9 +13,9 @@ __all__ = (
     'Coordinator', 'ExecutionOrder', 'Configuration', 'Parameters',
 
     # tags
-    'validate', 'Parameters', 'Configuration', 'Credentials', 'Shell', 'SubWorkflow', 'GlobalConfiguration', \
+    'Parameters', 'Configuration', 'Credentials', 'Shell', 'SubWorkflow', 'GlobalConfiguration', \
     'Email', 'IdentifierTooLongError',
 
     # builder
-    'workflow', 'coordinator', '_coordinator_submission_xml', '_workflow_submission_xml',
+    'workflow', 'coordinator',
 )
--- a/pyoozie/__init__.py
+++ b/pyoozie/__init__.py
@@ -4,7 +4,7 @@
 from pyoozie.coordinator import Coordinator, ExecutionOrder
 from pyoozie.tags import Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, \
     Email, IdentifierTooLongError
-from pyoozie.builder import workflow, coordinator
+from pyoozie.builder import WorkflowBuilder, CoordinatorBuilder
 
 __version__ = '0.0.0'
 
@@ -17,5 +17,5 @@ __all__ = (
     'Email', 'IdentifierTooLongError',
 
     # builder
-    'workflow', 'coordinator',
+    'WorkflowBuilder', 'CoordinatorBuilder',
 )
--- a/pyoozie/__init__.py
+++ b/pyoozie/__init__.py
@@ -1,5 +1,5 @@
 # Copyright (c) 2017 "Shopify inc." All rights reserved.
-# Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+# Use of this source code is governed by a MIT-style license that can be found in the LICENSE file.
 
 from pyoozie.coordinator import Coordinator, ExecutionOrder
 from pyoozie.tags import Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, Email
--- a/pyoozie/__init__.py
+++ b/pyoozie/__init__.py
@@ -1,4 +1,21 @@
 # Copyright (c) 2017 "Shopify inc." All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
+from pyoozie.coordinator import Coordinator, ExecutionOrder
+from pyoozie.tags import validate, Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, \
+    Email, IdentifierTooLongError
+from pyoozie.builder import workflow, coordinator, _coordinator_submission_xml, _workflow_submission_xml
+
 __version__ = '0.0.0'
+
+__all__ = (
+    # coordinator
+    'Coordinator', 'ExecutionOrder', 'Configuration', 'Parameters',
+
+    # tags
+    'validate', 'Parameters', 'Configuration', 'Credentials', 'Shell', 'SubWorkflow', 'GlobalConfiguration', \
+    'Email', 'IdentifierTooLongError',
+
+    # builder
+    'workflow', 'coordinator', '_coordinator_submission_xml', '_workflow_submission_xml',
+)
--- a/pyoozie/__init__.py
+++ b/pyoozie/__init__.py
@@ -2,8 +2,7 @@
 # Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
 from pyoozie.coordinator import Coordinator, ExecutionOrder
-from pyoozie.tags import Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, \
-    Email, IdentifierTooLongError
+from pyoozie.tags import Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, Email
 from pyoozie.builder import WorkflowBuilder, CoordinatorBuilder
 
 __version__ = '0.0.0'
@@ -13,8 +12,7 @@ __all__ = (
     'Coordinator', 'ExecutionOrder', 'Configuration', 'Parameters',
 
     # tags
-    'Parameters', 'Configuration', 'Credentials', 'Shell', 'SubWorkflow', 'GlobalConfiguration', \
-    'Email', 'IdentifierTooLongError',
+    'Parameters', 'Configuration', 'Credentials', 'Shell', 'SubWorkflow', 'GlobalConfiguration', 'Email',
 
     # builder
     'WorkflowBuilder', 'CoordinatorBuilder',
--- a/pyoozie/builder.py
+++ b/pyoozie/builder.py
@@ -39,14 +39,14 @@ class workflow(object):
     def add_action(self, name, action, action_on_error, kill_on_error='${wf:lastErrorNode()} - ${wf:id()}'):
         # Today you can't rename your action and you can only have one, but in the future you can add multiple
         # named actions
-        if self._action_name is None and self._action_payload is None and self._action_error is None and \
-           self._kill_message is None:
+        if any((self._action_name, self._action_payload, self._action_error, self._kill_message)):
+            raise NotImplementedError("Can only add one action in this version")
+        else:
             self._action_name = name
             self._action_payload = action
             self._action_error = action_on_error
             self._kill_message = kill_on_error
-        else:
-            raise NotImplementedError("Can only add one action in this version")
+
         return self
 
     def build(self, indent=False):
@@ -78,15 +78,18 @@ class workflow(object):
            action_name=self._action_name,
            name=self._name).strip()
 
-    def submit(self, oozie_url, hdfs_path, hadoop_user, hdfs_callback, timeout_in_seconds=None,
-               verbose=False, start=False, indent=False):
+    def submit_via(self, api, hdfs_path, hdfs_callback, start=False, indent=False):
         xml = self.build(indent=indent)
         hdfs_callback(hdfs_path, xml)
+        return api.jobs_submit_workflow(hdfs_path=hdfs_path, start=start)
+
+    def submit(self, oozie_url, hdfs_path, hadoop_user, hdfs_callback, timeout_in_seconds=None,
+               verbose=False, start=False, indent=False):
         # TODO create Oozie API and submit
         from mock import Mock
         OozieAPI = Mock()
         api = OozieAPI(url=oozie_url, user=hadoop_user, timeout=timeout_in_seconds, verbose=verbose)
-        api.jobs_submit_workflow(hdfs_path=hdfs_path, start=start)
+        self.submit_via(api=api, hdfs_path=hdfs_path, hdfs_callback=hdfs_callback, start=start, indent=indent)
         raise NotImplementedError()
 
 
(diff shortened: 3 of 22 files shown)


Makefile
```diff
@@ -24,5 +24,6 @@ autopep8:
 lint:
 		@echo 'Linting...'
 		@pylint --rcfile=pylintrc pyoozie tests
+		@flake8
```
cfournie: Let's use both linters

tests/utils.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+from __future__ import unicode_literals, print_function
+
+import xmltodict
+
+
+def xml_to_dict_unordered(xml):
```
cfournie: This replaces the entire pxxmltest library (it's error-reporting isn't as nice and it only tests entire docs, thanks to [`xmltodict`](https://github.com/martinblech/xmltodict) it's simple; thanks @honkfestival for finding this).

pyoozie/builder.py
```diff
@@ -0,0 +1,135 @@
...
+           name=self._name).strip()
+
+    def submit(self, oozie_url, hdfs_path, hadoop_user, hdfs_callback, timeout_in_seconds=None,
+               verbose=False, start=False, indent=False):
+        xml = self.build(indent=indent)
+        hdfs_callback(hdfs_path, xml)
+        # TODO create Oozie API and submit
+        from mock import Mock
+        OozieAPI = Mock()
+        api = OozieAPI(url=oozie_url, user=hadoop_user, timeout=timeout_in_seconds, verbose=verbose)
```
kmtaylor-github: Near term I think it'd make more sense to pass in a preconstructed `OozieAPI`, since the likely usecase is submitting multiple objects. Alternately, don't directly link these two at all (i.e. have the builder write to hdfs and stop).

pyoozie/builder.py
```diff
@@ -0,0 +1,135 @@
...
+        api.jobs_submit_workflow(hdfs_path=hdfs_path, start=start)
+        raise NotImplementedError()
+
+
+class coordinator(object):
+
+    def __init__(self, name, workflow, frequency_in_minutes, start, end=None, timezone=None,
+                 workflow_configuration=None, timeout_in_minutes=None, concurrency=None, execution_order=None,
+                 throttle=None, parameters=None):
```
kmtaylor-github: Why not builder-pattern all of this? Something like:
```
coordinator = coordinator('my-flow', workflow)
coordinator.set_frequency(360)
coordinator.set_start(datetime.now())
```

pyoozie/coordinator.py
```diff
@@ -0,0 +1,102 @@
...
+
+    LAST_ONLY = 'LAST_ONLY'
+    # "When LAST_ONLY is set, an action that is WAITING or READY will be SKIPPED when the current time is past the
+    # next action's nominal time.  For example, suppose action 1 and 2 are both WAITING , the current time is
+    # 5:00pm, and action 2's nominal time is 5:10pm. In 10 minutes from now, at 5:10pm, action 1 will become
+    # SKIPPED, assuming it doesn't transition to SUBMITTED (or a terminal state) before then. Another way of
+    # thinking about this is to view it as similar to setting the timeout equal to the frequency, except that the
+    # SKIPPED status doesn't cause the coordinator job to eventually become DONEWITHERROR and can actually become
+    # SUCCEEDED (i.e. it's a "good" version of TIMEDOUT ). LAST_ONLY is useful if you want a recurring job, but do
+    # not actually care about the individual instances and just always want the latest action."
```
kmtaylor-github: Do we really need Oozie doc pasted here? (I know, was there in the original source too.)

pyoozie/tags.py
```diff
@@ -0,0 +1,317 @@
...
+    </xml_tag>
+    """
+
+    def __init__(self, xml_tag, attributes=None, values=None):
+        Xml.__init__(self, xml_tag)
+        if values:
+            dict.__init__(self, values)
+        else:
+            dict.__init__(self)
+        self.attributes = attributes if attributes else dict()
```
kmtaylor-github: Alternate: `self.attributes = attributes or dict()`. I'm ambivalent, but I've come to like the `or` pattern where it fits.

kmtaylor-github: ... and then multiplex with `dict()` vs. `{}`. Good thing there's only one way to do things in python.

pyoozie/tags.py
```diff
@@ -0,0 +1,317 @@
...
+                for xml_file in self.job_xml_files:
+                    with tag('job-xml'):
+                        doc.text(xml_file)
+            if self.configuration:
+                self.configuration._xml(doc, tag, text)
+
+        return doc
+
+    def __nonzero__(self):
+        return any(bool(x) for x in [
```
kmtaylor-github: Do we need the `bool`?

pyoozie/tags.py
```diff
@@ -0,0 +1,317 @@
...
+        self.body = body
+        self.cc = cc
+        self.bcc = bcc
+        self.content_type = content_type
+        self.attachments = attachments
+
+    def _xml(self, doc, tag, text):
+        def format_list(emails):
+            return ','.join(sorted(emails)) if hasattr(emails, '__iter__') and not \
+                isinstance(emails, str) else emails
```
kmtaylor-github: Given that you've rolled onto two lines, should this _really_ need to be a single statement?

pyoozie/builder.py
```diff
@@ -0,0 +1,135 @@
...
+    """Generate a Coordinator XML submission message to POST to Oozie."""
+    submission = Configuration(configuration)
+    submission.update({
+        'user.name': hadoop_user,
+        'oozie.coord.application.path': hdfs_path,
+    })
+    return submission.xml(indent)
+
+
+class workflow(object):
```
cfournie: Probably because I added [`invalid-name`](http://pylint-messages.wikidot.com/messages:c0103) to pylintrc's ignore

pyoozie/builder.py
```diff
@@ -0,0 +1,135 @@
...
+from pyoozie.tags import Configuration, Parameters
+
+
+def _workflow_submission_xml(hadoop_user, hdfs_path, configuration=None, indent=False):
+    """Generate a Workflow XML submission message to POST to Oozie."""
+    submission = Configuration(configuration)
+    submission.update({
+        'user.name': hadoop_user,
+        'oozie.wf.application.path': hdfs_path
+    })
```
cfournie: Yes; I wanted to force `hadoop_user` and `hdfs_path` to be the preferred way to specify `user.name` and `oozie.wf.application.path`. Should I warn the user if they specify an overlapping config?

pyoozie/coordinator.py
cfournie: Probably not, but maybe as a docstring? I do want to explain what these are.

pyoozie/builder.py
cfournie: This may demonstrate that the builder pattern in Python is unneeded; let's discuss IRL.

pyoozie/tags.py
```diff
@@ -270,7 +270,7 @@ def _xml(self, doc, tag, text):
         return doc
 
     def __nonzero__(self):
-        return any(bool(x) for x in [
+        return any(x for x in [
             self.job_tracker, self.name_node, self.job_xml_files,
             self.configuration
```
kmtaylor-github: ... so what's the point of the generator now? 🙂

pyoozie/builder.py
```diff
@@ -0,0 +1,104 @@
...
+        self._name = name
+        self._action_name = None
+        self._action_payload = None
+        self._action_error = None
+        self._kill_message = None
+
+    def add_action(self, name, action, action_on_error, kill_on_error='${wf:lastErrorNode()} - ${wf:id()}'):
+        # Today you can't rename your action and you can only have one, but in the future you can add multiple
+        # named actions
+        if any((self._action_name, self._action_payload, self._action_error, self._kill_message)):
```
kmtaylor-github: Ha. I considered suggesting this pattern as a replacement for the multiline `if`, but decided not to. I guess we agree 🙂

pyoozie/tags.py
```diff
@@ -0,0 +1,319 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+from __future__ import unicode_literals
+from abc import ABCMeta, abstractmethod
+import re
+import yattag
+
+
+MAX_IDENTIFIER_LENGTH = 39
+REGEX_IDENTIFIER = r'^[a-zA-Z_][\-_a-zA-Z0-9]{0,38}$'
```
kmtaylor-github: Should this be `...[\\-_...`?

cfournie: Not with the power of ["raw strings"](https://docs.python.org/3.4/reference/lexical_analysis.html#string-and-bytes-literals) 😄 

> Both string and bytes literals may optionally be prefixed with a letter 'r' or 'R'; such strings are called raw strings and treat backslashes as literal characters.

pyoozie/tags.py
```diff
@@ -0,0 +1,313 @@
...
+        self.attributes = attributes or {}
+
+    def _xml(self, doc, tag, text):
+        with tag(self.xml_tag, **self.attributes):
+            for name, value in sorted(self.items()):
+                with tag('property'):
+                    with tag('name'):
+                        doc.text('%s' % name)
+                    with tag('value'):
+                        doc.text('%s' % value if value is not None else '')
```
honkfestival: You specifically want to catch `False` here, right?

pyoozie/tags.py
```diff
@@ -0,0 +1,313 @@
...
+    "Oozie allows a global section to reduce the redundant job-tracker and name-node declarations for each action.
+    [...] The global section may contain the job-xml, configuration, job-tracker, or name-node that the user would
+    like to set for every action.  If a user then redefines one of these in a specific action node, Oozie will
+    update [sic] use the specific declaration instead of the global one for that action."
+
+    "The job-xml element, if present, must refer to a Hadoop JobConf job.xml file bundled in the workflow
+    application."
+    """
+
+    def __init__(self, job_tracker='', name_node='', job_xml_files=None, configuration=None):
```
honkfestival: Can `job_tracker` and `name_node` default to `None`?

pyoozie/builder.py
```diff
@@ -0,0 +1,104 @@
...
+from pyoozie.tags import Configuration
+
+
+def _workflow_submission_xml(username, workflow_xml_path, configuration=None, indent=False):
+    """Generate a Workflow XML submission message to POST to Oozie."""
+    submission = Configuration(configuration)
+    submission.update({
+        'user.name': username,
+        'oozie.wf.application.path': workflow_xml_path,
+    })
```
kmtaylor-github: > Should I warn the user if they specify an overlapping config? Or raise an exception?

That's not a bad idea, but also not that important.

tests/pyoozie/test_tags.py
```diff
@@ -0,0 +1,293 @@
...
+                <value>${queueName}</value>
+            </property>
+        </configuration>
+    </global>
+    ''') == xml_to_dict_unordered(actual)
+
+
+def test_email():
+    actual = Email(
+        to='mrt@theateam.com',
```
honkfestival: Please use `example.com` email addresses everywhere in tests.

While real email addresses are more fun to read, using them runs the risk of leaking and/or spamming. `example.com` is [reserved by the IANA](http://www.iana.org/domains/reserved) specifically for illustrative purposes.

(I know you're not actually _sending_ anything here, but it's a good habit because someone else might.)

pyoozie/coordinator.py
```diff
@@ -0,0 +1,88 @@
...
+    LIFO = 'LIFO'
+    LAST_ONLY = 'LAST_ONLY'
+    NONE = 'NONE'
+
+    def __str__(self):
+        return self.value
+
+
+def format_datetime(value):
+    return value.strftime('%Y-%m-%dT%H:%MZ')
```
kmtaylor-github: Oozie's time grain is `minute`. You can probably specify seconds, but they'd be ignored.
From https://oozie.apache.org/docs/4.1.0/CoordinatorFunctionalSpec.html#a4._Datetime_Frequency_and_Time-Period_Representation

> If the Oozie processing timezone is UTC , all datetime values are always in UTC down to a minute precision, 'YYYY-MM-DDTHH:mmZ'.

pyoozie/coordinator.py
```diff
@@ -0,0 +1,88 @@
...
+
+class Coordinator(Xml):
+
+    def __init__(self, name, workflow_app_path, frequency, start, end=None, timezone=None,
+                 workflow_configuration=None, timeout=None, concurrency=None, execution_order=None, throttle=None,
+                 parameters=None):
+        super(Coordinator, self).__init__('coordinator-app')
+        # Compose and validate dates/frequencies
+        if end is None:
+            end = start + timedelta(days=100 * 365.24)
```
kmtaylor-github: Boo for Python not supporting `timedelta(years=100)`

pyoozie/builder.py
```diff
@@ -0,0 +1,104 @@
...
+            self._action_error = action_on_error
+            self._kill_message = kill_on_error
+
+        return self
+
+    def build(self, indent=False):
+        def format_xml(xml):
+            xml = xml.replace("<?xml version='1.0' encoding='UTF-8'?>", '')
+            return '\n'.join([(' ' * 8) + line for line in xml.strip().split('\n')])
+        return '''
```
cfournie: > Also, if we start having too many of these, we should consider using templates.

This inline template hopefully be a short-lived exception.

pyoozie/tags.py
```diff
@@ -0,0 +1,313 @@
...
+# Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
+from __future__ import unicode_literals
+from abc import ABCMeta, abstractmethod
+import re
+import yattag
+
+
+MAX_IDENTIFIER_LENGTH = 39
+REGEX_IDENTIFIER = r'^[a-zA-Z_][\-_a-zA-Z0-9]{0,38}$'
+COMPILED_REGEX_IDENTIFIER = re.compile(REGEX_IDENTIFIER)
```
honkfestival: ugh actually never mind; I don't know what I was thinking

pyoozie/__init__.py
```diff
@@ -1,4 +1,21 @@
 # Copyright (c) 2017 "Shopify inc." All rights reserved.
 # Use of this source code is governed by a BSD-style license that can be found in the LICENSE file.
 
+from pyoozie.coordinator import Coordinator, ExecutionOrder
+from pyoozie.tags import Parameters, Configuration, Credentials, Shell, SubWorkflow, GlobalConfiguration, \
+    Email, IdentifierTooLongError
+from pyoozie.builder import WorkflowBuilder, CoordinatorBuilder
+
 __version__ = '0.0.0'
```
cfournie: When should we update this; on every intended release? Every PR would become a bit tedious.

honkfestival: Not every PR, it just seems like `0.0.1` is a better starting place.

pyoozie/tags.py
cfournie: Correct; I want to be able to serialize the boolean `False` as the string `False`.

tests/pyoozie/test_tags.py
cfournie: I did not know that, and yeah, makes sense to prevent accidental spam.

tests/pyoozie/test_tags.py
```diff
@@ -237,56 +233,56 @@ def test_global_configuration():
...
         <body>Do you need more?</body>
     </email>
     ''') == xml_to_dict_unordered(actual)
 
     actual = Email(
-        to='mrt@theateam.com',
+        to='mrt@example.com',
         subject='Chains',
         body='Do you need more?',
         cc='ateam@ateam.com',
```
kmtaylor-github: Should probably hit the `cc`s too for consistency. 
Subdomains for colour maybe? (e.g. `mrt@theateam.example.com`) I'd be a shame to ruin perfectly amusing testcases.

honkfestival: I'm down with subdomains for colour!

PR #4, opened 2017-02-13
Add Oozie API

kmtaylor-github:
Add an interface layer on top of the Oozie REST API to query, submit, and manage workflow/coordinator artifacts.


pylintrc
```diff
@@ -47,7 +47,9 @@ load-plugins=
 # D102 Missing docstring in public method
 # D105 Missing docstring in magic method
 disable=C0111,D101,D102,D105,too-few-public-methods,too-many-instance-attributes,too-many-arguments,
- protected-access,invalid-name,redefined-outer-name,fixme
+ protected-access,invalid-name,redefined-outer-name,no-self-use,too-many-public-methods,too-many-lines,
+ no-member,access-member-before-definition,attribute-defined-outside-init,fixme
```
kmtaylor-github: I'm not thrilled about disabling some of these, but the dynamic nature of the code in `model.py` defeats many of the standard checks.

pylintrc
```diff
@@ -47,7 +47,9 @@ load-plugins=
 # D102 Missing docstring in public method
 # D105 Missing docstring in magic method
 disable=C0111,D101,D102,D105,too-few-public-methods,too-many-instance-attributes,too-many-arguments,
- protected-access,invalid-name,redefined-outer-name,fixme
+ protected-access,invalid-name,redefined-outer-name,no-self-use,too-many-public-methods,too-many-lines,
+ no-member,access-member-before-definition,attribute-defined-outside-init,duplicate-code,fixme
```
kmtaylor-github: We might be able to re-add some of these, but it'll require changing the structure of the code.
The dynamic nature of `model.py` confuses the linter.

pylintrc
```diff
@@ -101,7 +105,10 @@ ignore-mixin-members=yes
 
 # List of classes names for which member attributes should not be checked
 # (useful for classes with attributes dynamically set).
-ignored-classes=
+ignored-classes=pyoozie.model._OozieArtifact,
+ pyoozie.model._StatusValue, pyoozie.model._OozieArtifact.StatusEnum,
+ pyoozie.model.Coordinator, pyoozie.model.CoordinatorAction,
+ pyoozie.model.Workflow, pyoozie.model.WorkflowAction,
```
kmtaylor-github: Listing classes here prevents _some_ of the lint errors around `no-member` etc. -- but not all.
The `generated-members=` list below should also do this, but it doesn't appear to work properly with wildcards (despite the changelog entry promising exactly that).

pyoozie/model.py
```diff
@@ -39,61 +39,77 @@ def parse_workflow_id(string):
...
-    coord_id, action = parse_coordinator_id(job_id)
-    if coord_id and not action:
-        return job_id
-    raise OozieException.parse_error("Invalid coordinator id: {}".format(job_id))
+    if job_id:
+        coord_id, action = parse_coordinator_id(job_id)
+        if coord_id and not action:
+            return job_id
+        raise OozieException.parse_error("Invalid coordinator id: {}".format(job_id))
+    return None
```
kmtaylor-github: You'll see this pattern repeated below. Details in the `__init__`...

pyoozie/model.py
```diff
@@ -159,38 +170,37 @@ def suspended(cls):
             return [status for status in cls if status.is_suspended()]
 
         def is_unknown(self):
-            return self.value.status_id == 0
+            return self._value_.status_id == 0
```
kmtaylor-github: Linter won't accept `self.value` here in python3 -- but seems content with `_value_`. WTF?

pyoozie/model.py
```diff
@@ -159,38 +170,37 @@ def suspended(cls):
...
         self._oozie_api = oozie_api
         self._parent = parent
-        self.toString = None
         details = dict(details)
-        for key, func in self.REQUIRED_KEYS.iteritems():
+        for key, func in self.REQUIRED_KEYS.items():
             value = details.pop(key, None)
             try:
-                parsed_value = func(self, value) if func and value is not None else value
+                parsed_value = func(self, value) if func else value
```
kmtaylor-github: Subtle change: always call the parse function, even for `None`. This allows me the opportunity to tweak the default value; e.g. `status` should default to `UNKNOWN` if unspecified.

pyoozie/model.py
```diff
@@ -377,51 +361,25 @@ class Status(_OozieArtifact.StatusEnum):
...
-        else:
-            self.actionNumber = action
-        if self.toString:
-            if self.id not in self.toString:
-                raise OozieException.parse_error("toString does not contain coordinator action ID")
-            if not self.status.is_unknown() and str(self.status) not in self.toString:
-                raise OozieException.parse_error("toString does not contain status")
-        else:
-            self.toString = 'CoordinatorAction name[{}] status[{}]'.format(self.id, self.status)
+        self.coordJobId = self.coordJobId or coord_id
```
kmtaylor-github: Code here is logically the same as above, but rewritten so the linter won't be confused by the dynamic initialization that occurs in the `__init__` method.

tests/pyoozie/test_oozie_api.py
```diff
@@ -18,8 +18,20 @@
...
 @pytest.fixture
+def oozie_config():
+    return {
+        'url': 'http://localhost:11000/oozie',
+        'user': 'oozie',
+        'timeout': 30,
+        'verbose': False,
+        'launcher_memory_in_mb': '5000',
+        'launcher_queue': 'test.ignore',
+    }
```
kmtaylor-github: Originally appeared in `conftest.py`.

kmtaylor-github:
Ping @cfournie

pyoozie/model.py
```diff
@@ -0,0 +1,580 @@
...
+
+
+def _status(status_id, is_active=False, is_running=False, is_suspendable=False, is_suspended=False):
+    if is_running and not is_active:
+        raise OozieException.parse_error("A running status implies active")
+    return _StatusValue(status_id, is_active, is_running, is_suspendable, is_suspended)
+
+
+_COORD_ID_RE = re.compile('^(?P<id>.*-C)(?:@(?P<action>[1-9][0-9]*))?$')
+_WORKFLOW_ID_RE = re.compile('^(?P<id>.*-W)(?:@(?P<action>.*))?$')
```
cfournie: For consistency I'll rename my regex constants to match these https://github.com/Shopify/pyoozie/issues/7

pyoozie/oozie_api.py
```diff
@@ -0,0 +1,561 @@
+from __future__ import unicode_literals
+
+import logging
+import requests
+import simplejson as json
```
cfournie: Any reason why we can't use the stdlib `json` module?

pyoozie/oozie_api.py
```diff
@@ -0,0 +1,561 @@
+from __future__ import unicode_literals
```
cfournie: Should we rename this module `api`? `pyoozie.oozie_api` seems redundant.

pyoozie/oozie_api.py
```diff
@@ -0,0 +1,561 @@
...
+from pyoozie.exceptions import OozieException
+from pyoozie.model import ArtifactType, Coordinator, CoordinatorAction, \
+    Workflow, WorkflowAction, parse_coordinator_id, parse_workflow_id
+from pyoozie.builder import _coordinator_submission_xml, _workflow_submission_xml
+
+logger = logging.getLogger('pyoozie.OozieAPI')
+logging.getLogger('requests').setLevel(logging.WARNING)
+
+
+class OozieAPI(object):
```
cfournie: Should we rename this `Api`? Or some other noun, e.g. `Session` as in `requests`? Or perhaps `Client` and wrap it in a factory called `client`?  The `Oozie` might be redundant, e.g.:

```
oozie = pyoozie.OozieAPI(url='https://...', user='oozie', timeout=10)
```

As opposed to:


```
oozie_client = pyoozie.client(url='https://...', user='oozie', timeout=10)
```

pyoozie/oozie_api.py
```diff
@@ -0,0 +1,561 @@
...
+            self._stats.update(response)
+        except requests.RequestException as err:
+            self._stats.update(response)
+            if self._verbose and response is not None:
+                logger.error(response.headers)
+            message = "Unable to contact Oozie server at {}".format(self._url)
+            raise OozieException.communication_error(message, err)
+        versions = json.loads(response.content)
+        if 2 not in versions:
+            message = "Oozie server at {} does not support API version 2 (supported: {})".format(self._url, versions)
```
cfournie: As an aside, is there anything that v2 gives us that v1 doesn't? Should we eventually add v2 support?

pyoozie/oozie_api.py
cfournie: As an aside, this file is really long, but I don't think we should split it up. The linter is wrong in this case.

pyoozie/model.py
```diff
@@ -0,0 +1,580 @@
...
+
+def parse_coordinator_id(string):
+    parts = _COORD_ID_RE.match(string) if string else None
+    coord_id = parts.group('id') if parts else None
+    action = parts.group('action') if parts else None
+    action = int(action) if action else None
+    return coord_id, action
+
+
+def parse_workflow_id(string):
```
cfournie: Should this and `parse_coordinator_id` be added to `pyoozie.__all__`?

pyoozie/model.py
```diff
@@ -0,0 +1,580 @@
...
+
+def _parse_workflow_status(_, status_string):
+    return Workflow.Status.parse(status_string)
+
+
+def _parse_workflow_action_status(_, status_string):
+    return WorkflowAction.Status.parse(status_string)
+
+
+class ArtifactType(Enum):
```
cfournie: Should this and all other public classes be added to `pyoozie.__all__`?

pyoozie/oozie_api.py
cfournie: Should this and/or a factor be added to `pyoozie.__all__`?

kmtaylor-github: I'll buy `OozieClient`, but I'm opposed to potentially ambiguous names like `API` or `Client` outside of very limited scopes (and even then, not a big fan).

pyoozie/oozie_api.py
kmtaylor-github: > As an aside, is there anything that v2 gives us that v1 doesn't?

Yes, but don't remember what. I'll look.

> Should we eventually add v2 support?

Do you mean v1 support?

cfournie: Yep, sorry, read that incorrectly. Is `v1` useful?

pyoozie/oozie_api.py
cfournie: Fair enough. How about `OozieClient` but a `client` factory?

pyoozie/oozie_api.py
kmtaylor-github: Looks like v2 was introduced for Oozie 4.0, so circa Aug 5, 2013. 
Not too concerned about backwards compatibility being relevant.

pyoozie/oozie_api.py
kmtaylor-github: What's the value of a factory over the constructor in this case?

pyoozie/oozie_api.py
```diff
@@ -0,0 +1,561 @@
...
+
+    def jobs_workflow_names(self, user=None):
+        jobs = self._jobs_query(ArtifactType.Workflow, user=user, details=False)
+        return set([job.appName for job in jobs])
+
+    def jobs_all_coordinators(self, name=None, user=None, limit=0):
+        return self._jobs_query(ArtifactType.Coordinator, name=name, user=user, limit=limit)
+
+    def jobs_all_active_coordinators(self, user=None):
+        return self._jobs_query(ArtifactType.Coordinator, status=Coordinator.Status.active(), user=user)
```
cfournie: Can we take `Coordinator.Status` out of `Coordinator` and name it `CoordinatorStatus`?

kmtaylor-github:
Argh! Using `requests.json()` seems to appease python 3.3-3.5, but the error returned by 3.4+ doesn't match the [documented one](http://docs.python-requests.org/en/master/user/quickstart/#json-response-content).
```
E           assert 'No JSON object could be decoded' in '/home/travis/build/Shopify/pyoozie/pyoozie/oozie_client.py:143: pyoozie.exceptions.OozieCommunicationException: Expecting value: line 1 column 1 (char 0)'
E            +  where '/home/travis/build/Shopify/pyoozie/pyoozie/oozie_client.py:143: pyoozie.exceptions.OozieCommunicationException: Expecting value: line 1 column 1 (char 0)' = str(<ExceptionInfo OozieCommunicationException tblen=2>)
```

Digging deeper, the error is a `ValueError` subclass `JSONDecodeError`, but the message is not as expected.

pyoozie/model.py
```diff
@@ -0,0 +1,587 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
```
cfournie: Could we move this module under `pyoozie/api/_model.py` to make it private and put it under `api` to reduce ambiguity?

pyoozie/oozie_client.py
```diff
@@ -0,0 +1,571 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
```
cfournie: Could we move this module to `pyoozie/api/_client.py` to make it private and put it under `api` to reduce ambiguity?

setup.py
```diff
@@ -31,8 +31,11 @@
     packages=['pyoozie'],
     install_requires=[
         'enum34>=0.9.23',
-        'yattag>=1.7.2',
+        'requests>=2.12.3',
         'setuptools>=0.9',
+        'six',
+        'untangle',
```
cfournie: Is there a minimum version of these that we want to support?

pyoozie/exceptions.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
```
cfournie: Could we move this module under `pyoozie/_exceptions.py` to make it private?

setup.py
kmtaylor-github: That's a fair question. What do we want our strategy to be? Pinning to the newest version at the time of writing is one answer, but it's pretty weak.

cfournie: Ideally if there is a lower bound we should specify that.

kmtaylor-github: Apparently both undecorated packages are over a year old in the their current rev (4yrs for untangle!); I'll tag that as a minimum.

PR #5, opened 2017-02-14
Add validation on action names

cfournie:
Oozie [workflow action names are identifiers](https://github.com/apache/oozie/blob/35db5b31fa0c69680de4fe12eabc6fde818b2c54/client/src/main/resources/oozie-workflow-0.5.xsd#L159), and identifiers must conform to [`([a-zA-Z_]([\-_a-zA-Z0-9])*){1,39}`](https://github.com/apache/oozie/blob/35db5b31fa0c69680de4fe12eabc6fde818b2c54/client/src/main/resources/oozie-workflow-0.5.xsd#L26). This PR adds this enforcement.

### DIFF
--- a/tests/pyoozie/test_builder.py
+++ b/tests/pyoozie/test_builder.py
@@ -134,6 +134,29 @@ def test_workflow_builder():
     actual_xml = builder.build()
     assert xml_to_dict_unordered(expected) == xml_to_dict_unordered(actual_xml)
 
+    # Does it throw an exception on a bad name?
+    with pytest.raises(AssertionError) as assertion_info:
+        WorkflowBuilder(
+            name='descriptive-name'
+        ).add_action(
+            name='Name with invalid characters',
+            action=Shell(exec_command='echo "test"'),
+            action_on_error=Email(to='person@example.com', subject='Error', body='A bad thing happened'),
+            kill_on_error='Failure message',
+        )
+    assert str(assertion_info.value) == \
+        "Identifier must match ^[a-zA-Z_][\\-_a-zA-Z0-9]{0,38}$, 'Name with invalid characters' does not"
+
+    # Does it raise an exception when you try to add multiple actions?
+    with pytest.raises(NotImplementedError) as assertion_info:
+        builder.add_action(
+            name='payload',
+            action=Shell(exec_command='echo "test"'),
+            action_on_error=Email(to='person@example.com', subject='Error', body='A bad thing happened'),
+            kill_on_error='Failure message',
+        )
+    assert str(assertion_info.value) == 'Can only add one action in this version'
+
 
 def test_coordinator_builder(coordinator_xml_with_controls, workflow_app_path):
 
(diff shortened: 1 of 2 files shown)


tests/pyoozie/test_builder.py
```diff
@@ -134,6 +134,29 @@ def test_workflow_builder():
...
+        ).add_action(
+            name='Name with invalid characters',
+            action=Shell(exec_command='echo "test"'),
+            action_on_error=Email(to='person@example.com', subject='Error', body='A bad thing happened'),
+            kill_on_error='Failure message',
+        )
+    assert str(assertion_info.value) == \
+        "Identifier must match ^[a-zA-Z_][\\-_a-zA-Z0-9]{0,38}$, 'Name with invalid characters' does not"
+
+    # Does it raise an exception when you try to add multiple actions?
```
kmtaylor-github: Bonus test? 🙂

cfournie: Ah, yeaaah. That should be a different PR.

tests/pyoozie/test_builder.py
```diff
@@ -134,6 +134,29 @@ def test_workflow_builder():
...
+        WorkflowBuilder(
+            name='descriptive-name'
+        ).add_action(
+            name='Name with invalid characters',
+            action=Shell(exec_command='echo "test"'),
+            action_on_error=Email(to='person@example.com', subject='Error', body='A bad thing happened'),
+            kill_on_error='Failure message',
+        )
+    assert str(assertion_info.value) == \
+        "Identifier must match ^[a-zA-Z_][\\-_a-zA-Z0-9]{0,38}$, 'Name with invalid characters' does not"
```
honkfestival: Isn't it [`{1,39}`](https://github.com/apache/oozie/blob/35db5b31fa0c69680de4fe12eabc6fde818b2c54/client/src/main/resources/oozie-workflow-0.5.xsd#L26)?

cfournie: Yep, note the round brackets in the schema regex (`([a-zA-Z_]([\-_a-zA-Z0-9])*){1,39}`) and the square brackets in the Python implementation.

kmtaylor-github: facepalm-icon why why why would you specify your regex like that?

cfournie: Do you want me to make it closer to the Oozie XML schema's regex?

kmtaylor-github: No, yours is perfectly sensible. Theirs is not.

honkfestival:
LGTM

tests/pyoozie/test_builder.py
cfournie: Upon closer inspection, Oozie may have intended to limit identifiers to 39 characters, but because to the `*` they do not. My regex follows the spirit of their regex, but not what it actually restricts.

Should we use their regex with no length limit? Or my more limited regex?

kmtaylor-github: Sigh... I thought it was bad, but apparently it's worse.

I'd vote for respecting the intended (?) 39 char limit to ensure we don't get surprised if they fix the schema in the future.

honkfestival: Yeah, the prudent choice here is the right one.

PR #6, opened 2017-02-14
Check that we only allow one action at present

cfournie:
Let's test to make sure that at present we prevent the API from allowing multiple actions to be added.


cfournie:
> Getting an odd sense of déjà vu...

Yeah, this was carved out of the other validation one; I'm trying to keep my PRs on topic.

honkfestival:
👍

PR #9, opened 2017-02-15
Fix missing test coverage in tags

cfournie:
This increases the line test coverage to 100% for the tags module.

review/ @kmtaylor-github @honkfestival

### DIFF
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -46,10 +46,7 @@ class XMLSerializable(object):
 
     @abstractmethod
     def _xml(self, doc, tag, text):
-        raise NotImplementedError
-
-    def __str__(self):
-        return self.xml_tag
+        raise NotImplementedError()
 
 
 class _PropertyList(XMLSerializable, dict):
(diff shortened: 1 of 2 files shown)


kmtaylor-github:
```
# pylint: disable=abstract-method
class MyTag(XMLSerializable):
   ...
E       TypeError: Can't instantiate abstract class MyTag with abstract methods _xml
```

Sigh...

cfournie:
Bah. Well I'll just call it another (erroneous) way then.

honkfestival:
Jumping through hoops like this just for 100% coverage doesn't add much value. Is it possible to disable code coverage for these bits instead?

cfournie:
> Jumping through hoops like this just for 100% coverage doesn't add much value

True, but these aren't really hoops, they're legitimate tests of API functionality. I want to warn users loudly if they do something that's currently unsupported and I didn't test env vars or capture output at all.

honkfestival:
> they're legitimate tests of API functionality

Gotcha. From the PR description, I thought you were just chasing the 100. :)

Issue #10, opened 2017-02-15
Reorganize to keep most elements at the top-level and specific elements in xml or api


cfournie:
Let's try to enable usage to look like:

```
import pyoozie
import pyoozie.xml
import pyoozie.api

workflow_app_path = '...'

coord_xml = pyoozie.CoordinatorBuilder(
    name='coordinator-name',
    workflow_xml_path=workflow_app_path
    secret_feature=pzoozie.xml.SecretFeature()).build()

client = pyoozie.OozieClient(
    url='https://..',
    user='oozie',
    advanced_feature=pyoozie.api.AdvancedFeature())

coord_path = '...'

hdfs_client = '...'
hdfs_client.write(path=coord_path, data=coord_xml)

coord_config = {
    'user.name': 'oozie',
}

coordinator = oozie_client.jobs_submit_coordinator(
    coord_path,
    coord_config)

print(coordinator.coord_job_id)
assert coordinator.status.is_active
```

cfournie:
We went in a different direction: flat top-level repos.

PR #11, opened 2017-02-15
Refactor XML generation submodules and API visibility

cfournie:
Addressed https://github.com/Shopify/pyoozie/issues/10 by:
  - Making all packages private except for:
    - `pyoozie` for commonly needed objects; and
    - `pyoozie.xml/api` for more advanced classes/functions 
  - Fixing tests to import from where functions/classes are defined

review/ @kmtaylor-github @JasonMWhite


cfournie:
Actually let's keep this simple and instead favour https://github.com/Shopify/pyoozie/pull/14

PR #15, opened 2017-02-27
Split string validation into names or identifiers and enforce

cfournie:
Testing with Oozie 4.1.0 shows that identifiers (e.g. workflow action names) need to follow the regex `^[a-zA-Z_][\\-_a-zA-Z0-9]{0,38}$`, however names (e.g. workflow and coordinator names) do not need to adhere to that regex (e.g. can contain spaces/punctuation) but are limited to 255 chars by Oozie's backing database and are limited to printable ASCII characters.

Creating a workflow with a really long action name results in:
```
org.apache.oozie.command.CommandException: E0724: Invalid node name, name [action-ppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppppp] must be 50 chars or less
        at org.apache.oozie.command.wf.SubmitXCommand.execute(SubmitXCommand.java:272)
```

Creating a workflow with a very long name results in:
```
2017-02-27 17:35:36,418 ERROR SubmitXCommand:538 - SERVER[oozie] USER[root] GROUP[-] TOKEN[] APP[aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa] JOB[0000019-170227162353067-oozie-root-W] ACTION[-] XException,
org.apache.oozie.command.CommandException: E0803: IO error, E0603: SQL error in operation, <openjpa-2.2.2-r422266:1468616 fatal store error> org.apache.openjpa.persistence.RollbackException: The transaction has been rolled back.  See the nested exceptions for details on the errors that occurred.
FailedObject: org.apache.oozie.WorkflowJobBean@27d6d59
...
Caused by: <openjpa-2.2.2-r422266:1468616 fatal general error> org.apache.openjpa.persistence.PersistenceException: A truncation error was encountered trying to shrink VARCHAR 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa&' to length 255.

### CODE REVIEW

File: pyoozie/tags.py [hunk 1]
```diff
@@ -5,13 +5,22 @@
 import re
 import yattag
 
-
-MAX_IDENTIFIER_LENGTH = 39
-REGEX_IDENTIFIER = r'^[a-zA-Z_][\-_a-zA-Z0-9]{0,38}$'
+MAX_NAME_LENGTH = 255
+MAX_IDENTIFIER_LENGTH = 50
+REGEX_IDENTIFIER = r'^[a-zA-Z_][\-_a-zA-Z0-9]{0,254}$'
```
kmtaylor-github: Why don't the regex and `MAX_` value match?

cfournie: Because oops.

cfournie:
@honkfestival @kmtaylor-github this should add the necessary safety for names and IDs while being more permissive than before.

pyoozie/tags.py
```diff
@@ -3,16 +3,31 @@
 from __future__ import unicode_literals
 from abc import ABCMeta, abstractmethod
 import re
+import string  # pylint: disable=deprecated-module
```
cfournie: Some functions are deprecated in [Python 2](https://docs.python.org/2/library/string.html#deprecated-string-functions) and [Python 3](https://docs.python.org/3/library/string.html) but the module isn't deprecated, [Pylint is just a bit heavy-handed](https://www.logilab.org/ticket/2481).

pyoozie/tags.py
```diff
@@ -3,16 +3,31 @@
...
-def _validate(identifier):
+def _validate_name(name):
+    assert len(name) <= MAX_NAME_LENGTH, \
+        "Name must be less than {max_length} chars long, '{name}' is {length}".format(
+            max_length=MAX_NAME_LENGTH,
+            name=name,
+            length=len(name))
+
+    assert COMPILED_REGEX_NAME.match(name), \
+        "Name must be comprised of printable ASCII characters, '{name}' is not".format(name=name)
```
kmtaylor-github: Is a regex really the best way to check this? 
Are all printable characters really desirable (e.g. `\n`)?

cfournie: > Are all printable characters really desirable (e.g. \n)?

Probably not. `string.ascii_letters` + `string.punctuation` + `string.digits` + `' '` (space) is probably better.

> Is a regex really the best way to check this?

We could encode as ASCII and then check that the char is within a certain integer range, but that's a little difficult to read.

kmtaylor-github: > but that's a little difficult to read.

So is a dynamically constructed regex: `r'[%(chars)s]{1,%(max_length)i}'`

pyoozie/tags.py
```diff
@@ -3,16 +3,30 @@
...
 from abc import ABCMeta, abstractmethod
 import re
+import string  # pylint: disable=deprecated-module
 import yattag
 
-
-MAX_IDENTIFIER_LENGTH = 39
-REGEX_IDENTIFIER = r'^[a-zA-Z_][\-_a-zA-Z0-9]{0,38}$'
+MAX_NAME_LENGTH = 255
+MAX_IDENTIFIER_LENGTH = 50
```
honkfestival: Where does 50 come from? Manual testing?

tests/pyoozie/test_builder.py
```diff
@@ -153,18 +154,42 @@ def test_workflow_builder(tmpdir):
             xml=actual_xml,
         ))
 
-    # Does it throw an exception on a bad name?
+    # Does it throw an exception on a bad workflow name?
```
honkfestival: Instead of commenting sections of the test, could you split these into smaller tests with descriptive names? e.g. `test_builder_raises_on_bad_workflow_name`

tests/pyoozie/test_tags.py
```diff
@@ -62,30 +63,71 @@ def expected_property_values_xml():
...
+        _validate_id(very_long_name)
+    assert str(assertion_info.value) == (
+        "Identifier must be less than {max_length} chars long, '{identifier}' is {length}"
+    ).format(
+        max_length=MAX_IDENTIFIER_LENGTH,
+        identifier=very_long_name,
+        length=len(very_long_name)
+    )
+
+    # Id that doesn't satisfy regex because of bad start char
```
honkfestival: Ditty here - each of these should be separate tests.

pyoozie/tags.py
cfournie: Yes, it's hard to see in the PR, but the exception states:
```
org.apache.oozie.command.CommandException: E0724: Invalid node name, name [action-...] must be 50 chars or less

```

PR #16, opened 2017-03-01
Use extras_require

cfournie:
For consistency, let's move all dependencies into `setup.py`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ clean:
 		rm -fr *.egg-info
 
 install:
-		pip install -e .[test]
+		pip install -e ".[test]"
 		test -d lib || mkdir lib
 		test -f lib/oozie-client-4.1.0.jar || \
 			curl http://central.maven.org/maven2/org/apache/oozie/oozie-client/4.1.0/oozie-client-4.1.0.jar -o lib/oozie-client-4.1.0.jar
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,7 @@ clean:
 		rm -fr *.egg-info
 
 install:
-		pip install -e .
-		pip install -r dev_requirements.txt
+		pip install -e .[test]
 		test -d lib || mkdir lib
 		test -f lib/oozie-client-4.1.0.jar || \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -8,8 +8,7 @@ clean:
 		rm -fr *.egg-info
 
 install:
-		pip install -e .
-		pip install -r dev_requirements.txt
+		pip install -e .[test]
```
kmtaylor-github: I assume this is `pip` syntax? Would it be safer to quote it (to prevent free regex globbing).

honkfestival: +1 for quoting

setup.py
```diff
@@ -32,11 +32,26 @@
...
+            'flake8',
+            'mock',
+            'pylint',
+            'pytest-cov',
+            'pytest-randomly',
+            'pytest>=2.7',
+            'requests-mock',
+            'xmltodict',
+        ],
+    },
```
honkfestival: Neat, TIL

Makefile
cfournie: Those are [requirement specifiers](http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies), whose syntax, in this case, is described in [PEP 508](https://www.python.org/dev/peps/pep-0508/) as  `extras` and are comma-delimited ids within square brackets which pip interprets case-insensitively and maps to entries in `setup.py`'s `extras_require`. Quotes would be superfluous.

kmtaylor-github: > Quotes would be superfluous.

Are you sure? Makefile lines are executed via a `sh` compatible shell. 

```
$ echo [test]
[test]
$ touch t
$ echo [test]
t
```

cfournie: Ahh, yes, we should quote that entire pip argument. I thought that you meant that we should quote `pip install -e .["test"]`.

Makefile
```diff
@@ -8,8 +8,7 @@ clean:
 		rm -fr *.egg-info
 
 install:
-		pip install -e .
-		pip install -r dev_requirements.txt
+		pip install -e ".[test]"
```
kmtaylor-github: `:+1:`

PR #17, opened 2017-03-09
Add shopify_python linter and adhere to import-modules-only rule

cfournie:
This PR:
  - Adds the `shopify_python` linter plugin to pylint
  - Fixes violations of the `import-modules-only` rule
  - Disables the `global-variable` rule for now

Some of the changes to satisfy the `import-modules-only` rule look silly; those situations indicate  areas where we should maybe refactor, e.g. `coordinator.Coordinator` should perhaps be `xml.Coordinator`.

### DIFF
--- a/pyoozie/builder.py
+++ b/pyoozie/builder.py
@@ -2,13 +2,13 @@
 # Use of this source code is governed by a MIT-style license that can be found in the LICENSE file.
 from __future__ import unicode_literals
 
-from pyoozie.coordinator import Coordinator
-from pyoozie.tags import Configuration, _validate_id, _validate_name
+from pyoozie import coordinator
+from pyoozie import tags
 
 
 def _workflow_submission_xml(username, workflow_xml_path, configuration=None, indent=False):
     """Generate a Workflow XML submission message to POST to Oozie."""
-    submission = Configuration(configuration)
+    submission = tags.Configuration(configuration)
     submission.update({
         'user.name': username,
         'oozie.wf.application.path': workflow_xml_path,
@@ -18,7 +18,7 @@ def _workflow_submission_xml(username, workflow_xml_path, configuration=None, in
 
 def _coordinator_submission_xml(username, coord_xml_path, configuration=None, indent=False):
     """Generate a Coordinator XML submission message to POST to Oozie."""
-    submission = Configuration(configuration)
+    submission = tags.Configuration(configuration)
     submission.update({
         'user.name': username,
         'oozie.coord.application.path': coord_xml_path,
@@ -30,7 +30,7 @@ class WorkflowBuilder(object):
 
     def __init__(self, name):
         # Initially, let's just use a static template and only one action payload and one action on error
-        self._name = _validate_name(name)
+        self._name = tags._validate_name(name)
         self._action_name = None
         self._action_payload = None
         self._action_error = None
@@ -42,7 +42,7 @@ class WorkflowBuilder(object):
         if any((self._action_name, self._action_payload, self._action_error, self._kill_message)):
             raise NotImplementedError("Can only add one action in this version")
         else:
-            self._action_name = _validate_id('action-' + name)
+            self._action_name = tags._validate_id('action-' + name)
             self._action_payload = action
             self._action_error = action_on_error
             self._kill_message = kill_on_error
@@ -85,7 +85,7 @@ class CoordinatorBuilder(object):
     def __init__(self, name, workflow_xml_path, frequency_in_minutes, start, end=None, timezone=None,
                  workflow_configuration=None, timeout_in_minutes=None, concurrency=None, execution_order=None,
                  throttle=None, parameters=None):
-        self._coordinator = Coordinator(
+        self._coordinator = coordinator.Coordinator(
             name=name,
             workflow_app_path=workflow_xml_path,
             frequency=frequency_in_minutes,
--- a/setup.py
+++ b/setup.py
@@ -50,8 +50,12 @@ setup(
             'pytest>=2.7',
             'requests-mock',
             'xmltodict',
+            'shopify_python==0.1.2',
         ],
     },
+    dependency_links=[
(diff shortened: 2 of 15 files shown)


setup.py
```diff
@@ -50,8 +50,12 @@
             'pytest>=2.7',
             'requests-mock',
             'xmltodict',
+            'shopify_python==0.1.2',
         ],
     },
+    dependency_links=[
+        'git+https://github.com/Shopify/shopify_python.git@v0.1.2#egg=shopify_python-0.1.2',
```
honkfestival: How is this different from e.g.

```
-e git+https://github.com/Shopify/shopify_python.git@v0.1.2#egg=shopify_python-0.1.2
```

?

tests/pyoozie/test_builder.py
```diff
@@ -230,5 +229,6 @@ def test_coordinator_builder(coordinator_xml_with_controls, workflow_app_path):
         })
 
     # Can it XML?
-    expected_xml = builder.build()
-    assert xml_to_dict_unordered(coordinator_xml_with_controls) == xml_to_dict_unordered(expected_xml)
+    expected_xml = coord_builder.build()
+    assert tests.utils.xml_to_dict_unordered(coordinator_xml_with_controls) == \
+        tests.utils.xml_to_dict_unordered(expected_xml)
```
honkfestival: This would be more readable if you assigned the LHS and RHS to locals and then asserted equality. (You could also make it marginally better by using `from test import utils`, but I wouldn't worry about that.)

tests/pyoozie/test_tags.py
```diff
@@ -148,28 +147,28 @@ def _xml(self, doc, tag, text):
 
 
 def test_parameters(expected_property_values, expected_property_values_xml):
-    actual = Parameters(expected_property_values).xml(indent=True)
+    actual = tags.Parameters(expected_property_values).xml(indent=True)
     expected = '<parameters>{xml}</parameters>'.format(xml=expected_property_values_xml)
-    assert xml_to_dict_unordered(expected) == xml_to_dict_unordered(actual)
+    assert tests.utils.xml_to_dict_unordered(expected) == tests.utils.xml_to_dict_unordered(actual)
```
honkfestival: Same comment here re: `from tests import utils` but it's a nitpick really.

tests/pyoozie/test_coordinator.py
```diff
@@ -1,16 +1,16 @@
...
-from pyoozie.coordinator import Coordinator, ExecutionOrder, Configuration, Parameters
-from pyoozie.tags import MAX_NAME_LENGTH
-from tests.utils import xml_to_dict_unordered
+from pyoozie import coordinator
+from pyoozie import tags
 
 
 def parse_datetime(string):
-    return datetime.strptime(string, '%Y-%m-%dT%H:%MZ')
+    return datetime.datetime.strptime(string, '%Y-%m-%dT%H:%MZ')
```
kmtaylor-github: Of all the symbol->module import changes, this is the only one that bugs me. 
Note to self: never fall into the temptation of naming a module and class the same thing.

cfournie: It is irksome. If it irks us too badly we could rename it `as stddatetime`, but I'd rather not bother and instead, as you pointed out, learn from the std lib's mistakes.

honkfestival: While painful, at least it's also obvious what's happening. It irks me much more to not know which `datetime` I'm dealing with.

setup.py
cfournie: `-e git` would be syntactically incorrect:
```
InvalidSchema: No connection adapters were found for '-e git+https://github.com/Shopify/shopify_python.git@v0.1.2'
```

These are strictly links, not installation instructions (e.g. `pip -e <dependency>` states that installation from version controlled directories should be via symlink and not file copying).

pyoozie/builder.py
```diff
@@ -30,7 +30,7 @@ class WorkflowBuilder(object):
 
     def __init__(self, name):
         # Initially, let's just use a static template and only one action payload and one action on error
-        self._name = _validate_name(name)
+        self._name = tags._validate_name(name)
```
cfournie: I think that you're right; it's probably useful for someone developing their own name-generator to know ahead of time which names would be unacceptable (https://github.com/Shopify/pyoozie/issues/18). Thought for a future PR.

tests/pyoozie/test_tags.py
cfournie: I wanted to avoid importing `utils`; it doesn't communicate what sort of utils that we're using. The term is too generic. I would like to emphasize that these are testing utilities as opposed to the many other utilities that we may one day import.

setup.py
cfournie: In the long-term we shouldn't use `dependency_links` because they're deprecated. We should instead deploy `shopify_python` to pypi or switch to using a `dev_requirements.txt` for our CI/test env (although separating dependencies into separate files makes it harder to find them).

honkfestival: > syntactically incorrect

Right sorry it should have been `-e git+git://github.com` (but I did use _e.g._ 😛 )

If `dependency_links` is deprecated, let's switch to using separate requirements files soon (doens't have to be this PR).

tests/pyoozie/test_coordinator.py
```diff
@@ -31,51 +31,52 @@ def coordinator_xml():
...
         }),
-        parameters=Parameters({
+        parameters=tags.Parameters({
             'throttle': 1
         }),
         **expected_coordinator_options
     ).xml()
-    assert xml_to_dict_unordered(coordinator_xml_with_controls) == xml_to_dict_unordered(actual)
+    assert tests.utils.xml_to_dict_unordered(coordinator_xml_with_controls) == \
+        tests.utils.xml_to_dict_unordered(actual)
```
honkfestival: sorry @cfournie we missed one here too

PR #19, opened 2017-03-13
Make client logger per-instance to fix global-var lint rule

cfournie:
Let's remove side-effects from importing `pyoozie.client` by removing the global logger var that's defined therein.

### DIFF
--- a/pyoozie/client.py
+++ b/pyoozie/client.py
@@ -10,10 +10,6 @@ from pyoozie import exceptions
 from pyoozie import model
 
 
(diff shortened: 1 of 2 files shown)


pyoozie/client.py
```diff
@@ -10,10 +10,6 @@
 from pyoozie import model
 
 
-logger = logging.getLogger('pyoozie.OozieClient')
-logging.getLogger('requests').setLevel(logging.WARNING)
```
cfournie: This is probably useful for debugging but it's an import side-effect and we should let library users define their own requests logging levels.

PR #20, opened 2017-03-14
Update shopify python to v0.2.0 and fix lint errors

cfournie:
This PR updates `shopify_python` from 0.1.2 to 0.2.0 and syncs `pylintrc` with `shopify_python` which brings with it a number of new linter rules checks. This PR also fixes lint errors by:
  - Removing multiple imports on a line
  - Moving a lot of code out of `try`'s
  - Increasing the max except block size

### DIFF
--- a/pyoozie/client.py
+++ b/pyoozie/client.py
@@ -109,22 +109,18 @@ class OozieClient(object):
 
     def _request(self, method, endpoint, content_type, content=None):
         response = None
+        url = '{}/v2/{}'.format(self._url, endpoint)
+
+        if self._verbose:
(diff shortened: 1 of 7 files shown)


pyoozie/client.py
```diff
@@ -310,29 +314,32 @@ def _coordinator_query(self, job_id, status=None, start=0, limit=0):
                 start = int(action)
                 limit = 1
 
+        def wrapped_get(uri):
```
JasonMWhite: This feels a little clunky, but I can't think of a better pattern right now.

PR #25, opened 2017-03-21
[WIP] Prototype workflow builder interface

cfournie:
Here's what I'd like to use as an API for the workflow builder that allows people to specify:
  - Actions that they want executed (and their dependencies or other metadata), and
  - Transformations of the workflow execution graph using those actions and the action metadata specified.

This would be the final PR merged to add the XML workflow generation feature, before this there will need to be PRs that:
  1. Add a basic `Workflow` tag/graph to `pyoozie.tags`; and
  2. Add some basic transformation functions (and the graph functions required to implement them).

For an initial review, I'd like design comments and questions.

### DIFF
--- a/pyoozie/builder.py
+++ b/pyoozie/builder.py
@@ -56,19 +56,23 @@ class WorkflowBuilder(object):
         self.__action_layout_kwargs = kwargs
 
     def add_action(self, name, action, depends_upon=None, **kwargs):
-        def create_action_name(name):
-            return tags.validate_xml_id('action-' + name)
 
         # Validate name
-        name = create_action_name(name)
+        name = tags.validate_xml_id('action-' + name)
         assert name not in self.__actions, "Cannot add an action with the same name (%(name)s) twice".format(
             name=name)
+        assert action not in self.__actions.values(), "Cannot add the same action (%(action)r) twice".format(
+            action=action
+        )
 
         # Compose and validate dependency names
         dependencies = set()
-        for dependency_name in depends_upon or {}:
-            dependencies.add(create_action_name(dependency_name))
-        self.__dependencies[name] = dependencies
+        for dependency in depends_upon or {}:
+            assert dependency in self.__actions.values(), (
+                "%(dependency)r required by %(action)r but hasn't been added"
+                .format(dependency=dependency, action=action))
+            dependencies.add(dependency)
+        self.__dependencies[action] = dependencies
 
         # Store actions
         self.__actions[name] = action
--- a/pyoozie/builder.py
+++ b/pyoozie/builder.py
@@ -3,6 +3,7 @@
 from __future__ import unicode_literals
 
 from pyoozie import tags
+from pyoozie import transforms
 
 
 def _workflow_submission_xml(username, workflow_xml_path, configuration=None, indent=False):
@@ -27,56 +28,82 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
 
 class WorkflowBuilder(object):
 
-    def __init__(self, name):
-        # Initially, let's just use a static template and only one action payload and one action on error
-        self._name = tags.validate_xml_name(name)
-        self._action_name = None
-        self._action_payload = None
-        self._action_error = None
-        self._kill_message = None
+    def __init__(self, name, parameters=None, configuration=None, credentials=None, job_tracker=None,
+                 name_node=None, job_xml_files=None, default_retry_max=None, default_retry_interval=None,
+                 default_retry_policy=None):
+        self.__workflow = tags.Workflow(
+            name=name,
+            parameters=parameters,
+            configuration=configuration,
+            credentials=credentials,
+            job_tracker=job_tracker,
+            name_node=name_node,
+            job_xml_files=job_xml_files,
+            default_retry_max=default_retry_max,
+            default_retry_interval=default_retry_interval,
+            default_retry_policy=default_retry_policy,
+        )
+        self.__action_layout_strategy = transforms.serial_layout
+        self.__action_layout_kwargs = {}
+        self.__actions = {}
+        self.__action_kwargs = {}
+        self.__dependencies = {}
+        self.__transformations = []
+        self.__transformation_kwargs = {}
 
-    def add_action(self, name, action, action_on_error, kill_on_error='${wf:lastErrorNode()} - ${wf:id()}'):
-        # Today you can't rename your action and you can only have one, but in the future you can add multiple
-        # named actions
-        if any((self._action_name, self._action_payload, self._action_error, self._kill_message)):
-            raise NotImplementedError("Can only add one action in this version")
-        else:
-            self._action_name = tags.validate_xml_id('action-' + name)
-            self._action_payload = action
-            self._action_error = action_on_error
-            self._kill_message = kill_on_error
+    def set_layout(self, fnc_layout, **kwargs):
+        self.__action_layout_strategy = fnc_layout
+        self.__action_layout_kwargs = kwargs
+
+    def add_action(self, name, action, depends_upon=None, **kwargs):
+        def create_action_name(name):
+            return tags.validate_xml_id('action-' + name)
+
+        # Validate name
+        name = create_action_name(name)
+        assert name not in self.__actions, "Cannot add an action with the same name (%(name)s) twice".format(
+            name=name)
+
+        # Compose and validate dependency names
+        dependencies = set()
+        for dependency_name in depends_upon or {}:
+            dependencies.add(create_action_name(dependency_name))
+        self.__dependencies[name] = dependencies
+
+        # Store actions
+        self.__actions[name] = action
+        self.__action_kwargs[name] = kwargs
 
         return self
 
+    def add_transform(self, fnc_transform, **kwargs):
+        assert fnc_transform not in self.__transformation_kwargs, (
+            "Cannot add a transformation function (%(fnc)r) twice".format(fnc=fnc_transform))
+        self.__transformation_kwargs[fnc_transform] = kwargs
+        self.__transformations.append(fnc_transform)
+        return self
+
     def build(self, indent=False):
-        def format_xml(xml):
-            xml = xml.replace("<?xml version='1.0' encoding='UTF-8'?>", '')
-            return '\n'.join([(' ' * 8) + line for line in xml.strip().split('\n')])
-        return '''
-<?xml version="1.0" encoding="UTF-8"?>
-<workflow-app xmlns="uri:oozie:workflow:0.5"
-              name="{name}">
-    <start to="{action_name}" />
-    <action name="{action_name}">
-{action_payload_xml}
-        <ok to="end" />
-        <error to="action-error" />
-    </action>
-    <action name="action-error">
-{action_error_xml}
-        <ok to="kill" />
-        <error to="kill" />
-    </action>
-    <kill name="kill">
-        <message>{kill_message}</message>
-    </kill>
-    <end name="end" />
-</workflow-app>
-'''.format(action_payload_xml=format_xml(self._action_payload.xml(indent=indent)),
-           action_error_xml=format_xml(self._action_error.xml(indent=indent)),
-           kill_message=self._kill_message,
-           action_name=self._action_name,
-           name=self._name).strip()
+        env = {
+            'actions': self.__actions,
+            'action_kwargs': self.__action_kwargs,
+            'dependencies': self.__dependencies,
+            'workflow': self.__workflow,
+        }
+
+        # Ensure that this operation can only be called once
+        assert self.__workflow, "Workflow already built"
+        self.__workflow = None
+
+        # Layout the action nodes provided
+        env = self.__action_layout_strategy(env, **self.__action_layout_kwargs)
+
+        # Manipulate those action nodes (and potentially add others)
+        for transformation in self.__transformations:
+            transformation_kwargs = self.__transformation_kwargs[transformation]
+            env = transformation(env, **transformation_kwargs)
+
+        return env['workflow'].xml(indent=indent)
 
 
 class CoordinatorBuilder(object):
--- a/pyoozie/transforms.py
+++ b/pyoozie/transforms.py
@@ -1,0 +1,60 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
+# Use of this source code is governed by a MIT-style license that can be found in the LICENSE file.
+import six
+
+
+MESSAGE_ON_KILL = '${wf:lastErrorNode()} - ${wf:id()}'
+
+
+def serial_layout(env, error_to_kill=MESSAGE_ON_KILL):
+    """
+    TODO
+    1. Validate that all dependencies are provided
+    2. Construct a linear sequence of actions by performing these steps:
+      2a. Add all actions with no dependencies
+      2b. For all actions added, remove them as dependencies for un-added actions
+      3c. If there are no actions left to add, return, else go to 2a
+    3. If `error_to_kill` add a kill node and route all errors to it else route all errors to the next node
+    """
+    assert 'workflow' in env
+    assert 'dependencies' in env
+    assert error_to_kill is None or isinstance(error_to_kill, six.string_types)
+    return env
+
+
+def fork_join_layout(env):
+    """
+    TODO
+    """
+    return env
+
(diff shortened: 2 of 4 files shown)


pyoozie/builder.py
```diff
@@ -27,56 +28,77 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
-        if any((self._action_name, self._action_payload, self._action_error, self._kill_message)):
-            raise NotImplementedError("Can only add one action in this version")
-        else:
-            self._action_name = tags.validate_xml_id('action-' + name)
-            self._action_payload = action
-            self._action_error = action_on_error
-            self._kill_message = kill_on_error
+    def __init__(self, name, action_layout_strategy=transforms.serial_layout, parameters=None,
+                 configuration=None, credentials=None, job_tracker=None, name_node=None, job_xml_files=None,
+                 default_retry_max=None, default_retry_interval=None, default_retry_policy=None):
```
cfournie: Let's accept all of the fields that a workflow tag needs and also an `action_layout_strategy` function (e.g. do we want our nodes arranged serially? Do we want forks/joins according to some alg?).

pyoozie/builder.py
```diff
@@ -27,56 +28,77 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
 
 class WorkflowBuilder(object):
 
-    def __init__(self, name):
-        # Initially, let's just use a static template and only one action payload and one action on error
-        self._name = tags.validate_xml_name(name)
```
cfournie: Move validation into `tags.Workflow`'s constructor.

pyoozie/builder.py
```diff
@@ -27,56 +28,77 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+            default_retry_policy=default_retry_policy,
+        )
+        self.__action_layout_strategy = action_layout_strategy
+        self.__actions = {}
+        self.__action_kwargs = {}
+        self.__dependencies = {}
+        self.__transformations = []
+        self.__transformation_kwargs = {}
+
+    def add_action(self, name, action, depends_upon=None, **kwargs):
```
cfournie: Add actions and accept:
  - `name`, `action`, and `depends_upon` for the `action_layout_strategy`; and
  - transformation-specific metadata per-action via `kwargs`.

pyoozie/builder.py
```diff
@@ -27,56 +28,77 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+            dependencies.add(create_action_name(dependency_name))
+        self.__dependencies[name] = dependencies
+
+        # Store actions
+        self.__actions[name] = action
+        self.__action_kwargs[name] = kwargs
 
         return self
 
+    def add_transform(self, fnc_transform, **kwargs):
```
cfournie: Add a transformation function to be run on the graph of nodes within the workflow and any arguments that it should take.

pyoozie/builder.py
```diff
@@ -27,56 +28,77 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
-            raise NotImplementedError("Can only add one action in this version")
-        else:
-            self._action_name = tags.validate_xml_id('action-' + name)
-            self._action_payload = action
-            self._action_error = action_on_error
-            self._kill_message = kill_on_error
+    def __init__(self, name, action_layout_strategy=transforms.serial_layout, parameters=None,
+                 configuration=None, credentials=None, job_tracker=None, name_node=None, job_xml_files=None,
+                 default_retry_max=None, default_retry_interval=None, default_retry_policy=None):
+        self.__workflow = tags.Workflow(
```
cfournie: Let's start with a simple workflow graph of a `start` node that unconditionally ends at an `end` node.

pyoozie/builder.py
```diff
@@ -27,56 +28,77 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+            'dependencies': self.__dependencies,
+            'workflow': self.__workflow,
+        }
+
+        # Ensure that this operation can only be called once
+        assert self.__workflow, "Workflow already built"
+        self.__workflow = None
+
+        # Layout the action nodes provided
+        env = self.__action_layout_strategy(env)
```
cfournie: Apply our layout strategy, so our `start->end` graph has all actions added (e.g. `start->action1->end`)

pyoozie/builder.py
```diff
@@ -27,56 +28,77 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+
+        # Manipulate those action nodes (and potentially add others)
+        for transformation in self.__transformations:
+            transformation_kwargs = self.__transformation_kwargs[transformation]
+            env = transformation(env, **transformation_kwargs)
```
cfournie: Apply transformations in order to add features such as:
  - A final success notification (`transforms.final_action`); and
  - Notifications actions on error (`transforms. add_actions_on_error`).

tests/pyoozie/test_builder.py
```diff
@@ -130,7 +119,37 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
...
+                                   body='A bad thing happened while loading'),
+        kill_on_error='Failure message on loading',
+        depends_upon=('transform',),
+    ).add_transform(
+        transforms.final_action,
+        action=tags.Email(to='person@example.com', subject='Success', body='ETL succeeded'),
+    ).add_transform(
+        transforms.add_actions_on_error
+    )
+
```
cfournie: Above is what the interface would look like to a user who wanted a serial workflow with notifications on an error.

pyoozie/builder.py
```diff
@@ -27,56 +28,82 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+        self.__action_kwargs = {}
+        self.__dependencies = {}
+        self.__transformations = []
+        self.__transformation_kwargs = {}
+
+    def set_layout(self, fnc_layout, **kwargs):
+        self.__action_layout_strategy = fnc_layout
+        self.__action_layout_kwargs = kwargs
+
+    def add_action(self, name, action, depends_upon=None, **kwargs):
```
cfournie: Add actions and accept:
  - `name`, `action`, and `depends_upon` for the `action_layout_strategy`; and
  - transformation-specific metadata per-action via `kwargs`.

pyoozie/builder.py
```diff
@@ -27,56 +28,82 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+            dependencies.add(create_action_name(dependency_name))
+        self.__dependencies[name] = dependencies
 
+        # Store actions
+        self.__actions[name] = action
+        self.__action_kwargs[name] = kwargs
+
+        return self
+
+    def add_transform(self, fnc_transform, **kwargs):
```
cfournie: Add a transformation function to be run on the graph of nodes within the workflow and any arguments that it should take.

pyoozie/builder.py
```diff
@@ -27,56 +28,82 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+        self.__action_layout_kwargs = {}
+        self.__actions = {}
+        self.__action_kwargs = {}
+        self.__dependencies = {}
+        self.__transformations = []
+        self.__transformation_kwargs = {}
+
+    def set_layout(self, fnc_layout, **kwargs):
+        self.__action_layout_strategy = fnc_layout
+        self.__action_layout_kwargs = kwargs
```
cfournie: Allow setting an `action_layout_strategy` function (e.g. do we want our nodes arranged serially? Do we want forks/joins according to some alg?).

pyoozie/builder.py
```diff
@@ -27,56 +28,82 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
-        if any((self._action_name, self._action_payload, self._action_error, self._kill_message)):
-            raise NotImplementedError("Can only add one action in this version")
-        else:
-            self._action_name = tags.validate_xml_id('action-' + name)
-            self._action_payload = action
-            self._action_error = action_on_error
-            self._kill_message = kill_on_error
+    def __init__(self, name, parameters=None, configuration=None, credentials=None, job_tracker=None,
+                 name_node=None, job_xml_files=None, default_retry_max=None, default_retry_interval=None,
+                 default_retry_policy=None):
```
cfournie: Let's accept all of the fields that a workflow tag needs.

pyoozie/builder.py
```diff
@@ -27,56 +28,82 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+        self.__workflow = tags.Workflow(
```
cfournie: Let's start with a simple workflow graph of a `start` node that unconditionally ends at an `end` node.

pyoozie/builder.py
```diff
@@ -27,56 +28,82 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
+            'dependencies': self.__dependencies,
+            'workflow': self.__workflow,
+        }
+
+        # Ensure that this operation can only be called once
+        assert self.__workflow, "Workflow already built"
+        self.__workflow = None
+
+        # Layout the action nodes provided
+        env = self.__action_layout_strategy(env, **self.__action_layout_kwargs)
```
cfournie: Apply our layout strategy, so our `start->end` graph has all actions added (e.g. `start->action1->end`)

pyoozie/transforms.py
```diff
@@ -0,0 +1,60 @@
...
+    Relies upon action kwargs to add on-failure actions and kill nodes:
+      - action_on_error (action)
+      - kill_on_error (message, optional)
+    Validate kwargs and then then for each action in env['action']:
+      1. Create/select a kill node to end at (if `kill_on_error`)
+      2. Set error transition to go to `action_on_error`
+      3. Set ok/error for `action_on_error` to the kill node chosen OR to the original action's error node
+
+    Use cases:
+      - Notifying on an error and then optionally going to kill
```
kmtaylor-github: ...optionally going to kill? What other choice is there, and how would we specify it?

tests/pyoozie/test_builder.py
```diff
@@ -130,7 +119,39 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
...
+def test_workflow_builder(tmpdir):
+
+    workflow_builder = builder.WorkflowBuilder(
+        name='descriptive-name',
+        job_tracker='job-tracker',
+        name_node='name-node',
+    ).add_action(
+        name='extract',
+        action=tags.Shell(exec_command='echo', arguments=["'Extract data from an operational system'"]),
+        action_on_error=tags.Email(to='person@example.com', subject='Error',
```
kmtaylor-github: So the `action_on_error` is an action node; there's no way to specify an action by name?

pyoozie/transforms.py
```diff
@@ -0,0 +1,60 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
+# Use of this source code is governed by a MIT-style license that can be found in the LICENSE file.
+import six
+
+
+MESSAGE_ON_KILL = '${wf:lastErrorNode()} - ${wf:id()}'
+
+
+def serial_layout(env, error_to_kill=MESSAGE_ON_KILL):
```
cfournie:
```
  A
 / \
B   C
```

Could be rearranged as `A -> B -> C`

pyoozie/transforms.py
cfournie: If `kill_on_error` is specified, go to a kill node with that text. If it's not specified, go to where the action that we're notifying on was originally going on error (e.g. to another action, to `end`)

tests/pyoozie/test_builder.py
cfournie: I'd like to keep users as oblivious to names as I can; if you wanted to share the same action on error between actions we could give the same action_on_error instance to multiple actions.

pyoozie/transforms.py
cfournie: As long as we don't support cycles, we should be able to represent every dependency tree as a sub-optimal chain.

tests/pyoozie/test_builder.py
honkfestival: You could make the same argument of dependency instances too. If we support specifying actions by name for dependencies, we should also support it for errors.

honkfestival:
I think this would be simpler if you remove the concept of a layout entirely and only support actions with dependencies.

Does it really matter how the actions are laid out if all the dependencies are met? A [topological sort](https://en.wikipedia.org/wiki/Topological_sorting) on actions with their dependencies as edges should suffice.

cfournie:
> I think this would be simpler if you remove the concept of a layout entirely and only support actions with dependencies.

At least one action must have no dependencies (as an initial action). I'm trying to offer a way for someone to specify:
  - work actions (i.e. actions that produce something that you care about between which there are dependencies that manifest as edges specified by a layout); and
  - decorator actions (i.e. actions, edges, and other control-flow nodes such as fork/join/kill that determine how to act in error situations including the insertion of notification actions).

> Does it really matter how the actions are laid out if all the dependencies are met? A topological sort on actions with their dependencies as edges should suffice.

There are two main ways that we can lay out work actions:
  - Sequentially using a topological sort as you suggested; or
  - As a fork-join DAG that introduces parallel execution at times.

The fork-join DAG could be lain out in a wide variety of ways; we'll provide a naive implementation that relies upon just dependency information to create the graph but you may want to define your own strategy which may:
  - Take into account median job runtime to rearrange the workflow to minimize the critical path;
  - Multiply schedule the same job within a workflow and use external synchronization to minimize the critical path.
Both are strategies that we've discussed investigating.

tests/pyoozie/test_builder.py
cfournie: Hmm, maybe we shouldn't support specifying by name at all.

honkfestival: > Hmm, maybe we shouldn't support specifying by name at all.

I'd be down with that too. 👍

tests/pyoozie/test_builder.py
```diff
@@ -130,12 +119,55 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
...
 
-def test_workflow_builder(tmpdir, workflow_builder):
-    with open('tests/data/workflow.xml', 'r') as fh:
-        expected_xml = fh.read()
+@pytest.mark.xfail
+def test_workflow_builder(tmpdir):
+
+    extract_action = tags.Shell(exec_command='echo', arguments=["'Extract data from an operational system'"])
+    transform_action = tags.Shell(exec_command='echo', arguments=["'Transform data'"])
+    load_action = tags.Shell(exec_command='echo', arguments=["'Load data into a database'"])
```
cfournie: Let's define our work actions up front

tests/pyoozie/test_builder.py
```diff
@@ -130,12 +119,55 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
...
+        action_on_error=tags.Email(to='person@example.com', subject='Error',
+                                   body='A bad thing happened while extracting'),
+        kill_on_error='Failure message on extracting',
+    ).add_action(
+        name='transform',
+        action=transform_action,
+        action_on_error=tags.Email(to='person@example.com', subject='Error',
+                                   body='A bad thing happened while transforming'),
+        kill_on_error='Failure message on transforming',
+        depends_upon=(extract_action,),
```
cfournie: Actions are added one at a time with the assumption that an dependencies have already been added (if not, an assertion is thrown).

honkfestival:
> Sequentially using a topological sort as you suggested; or

Using topological sort doesn't require that we run the jobs sequentially. Rosetta Code has an [example in Python](https://rosettacode.org/wiki/Topological_sort#Python) of how to sort so that you know which groups can be run in parallel.

> - Take into account median job runtime to rearrange the workflow to minimize the critical path;

Minimizing the critical path _necessarily_ means breaking up the DAG, which would mean breaking one flow into multiple flows. I don't see how manually specifying fork/join points would help (but I'm probably missing some context).

> - Multiply schedule the same job within a workflow and use external synchronization to minimize the critical path.

Sorry, but I don't understand this. Let's chat IRL to make sure I'm not missing something.

cfournie:
Talked IRL and I'll begin prototyping a different approach.

PR #26, opened 2017-03-22
[WIP] Add prototype scoping interface to defining a workflow

cfournie:
This is a prototype interface that would allow one to define the following fork/join graph and (with inherent fork/joins) with the ability to specify actions to perform on error at each "scope" (i.e. wrapped action, sequence of actions, or fork/join of actions).

The following graph doesn't show error handling nodes, but when you don't define error handling for a scope then we exit the current scope and execute the parent scope's error handling (much like exception handling works).

```
   S
   |
   A
   |
 Fork
 /   \
B     C
|     |
|     D
 \   /
 Join
   |
   E
```

Not intended to be merged.

### DIFF
--- a/pyoozie/workflow.py
+++ b/pyoozie/workflow.py
@@ -1,43 +1,81 @@
 # Copyright (c) 2017 "Shopify inc." All rights reserved.
 # Use of this source code is governed by a MIT-style license that can be found in the LICENSE file.
+import abc
+import copy
 import typing  # pylint: disable=unused-import
 
 from pyoozie import tags  # pylint: disable=unused-import
 
 
-class AbstractAction(object):
-    pass
+# TODO auto-generate Oozie compatible ids
+
+
+class AbstractIdentifiedNode(object):
+    """Workflow tags representing nodes in a workflow's execution, e.g. Kill, and Action"""
+
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, identifier):
+        # type: (AbstractAction) -> None
+        self.__identifier = identifier
+
+    def identifier(self):
+        return self.__identifier
+
+
+class AbstractAction(AbstractIdentifiedNode):
+    """Parent class of both Action and action collections"""
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, identifier=None, on_error=None):
+        # type: (str, AbstractIdentifiedNode) -> None
+        super(AbstractAction, self).__init__(identifier=identifier or 'auto-generated-id-here')
+        self.__on_error = copy.deepcopy(on_error)
+
+    def on_error(self):
+        return self.__on_error
 
 
 class Action(AbstractAction):
     """Concrete action to execute (takes an action tag, e.g. pyoozie.Shell, pyoozie.Email. etc.)"""
 
     def __init__(self, action_tag, on_error=None):
-        # type: (tags.Shell, AbstractAction) -> None
-        self.__action_tag = action_tag
-        self.__on_error = on_error
+        # type: (typing.Union[tags.Email, tags.Shell, tags.SubWorkflow], AbstractIdentifiedNode) -> None
+        super(Action, self).__init__(identifier='auto-generated', on_error=on_error)
+        self.action_tag = action_tag
 
 
-class Sequence(AbstractAction):
-    """Set of actions to execute sequentially (implemented by chaining actions and 'OK' transitions)"""
+class Serial(AbstractAction):
+    """Collection of actions to execute sequentially (implemented by chaining actions and 'OK' transitions)"""
 
     def __init__(self, *actions, **kwargs):
-        # type: (typing.Tuple[AbstractAction, ...], typing.Dict[str, AbstractAction]) -> None
-        self.__on_error = kwargs.get('on_error')
-        self.__actions = actions
+        # type: (typing.Tuple[AbstractAction, ...], typing.Dict[str, AbstractIdentifiedNode]) -> None
+        super(Serial, self).__init__(identifier='auto-generated', on_error=kwargs.get('on_error'))
+        self.actions = copy.deepcopy(actions)
 
 
-class Parallelization(AbstractAction):
+class Parallel(AbstractAction):
     """Set of actions to execute in parallel (implemented as fork/join tag pair)"""
 
     def __init__(self, *actions, **kwargs):
-        # type: (typing.Tuple[AbstractAction, ...], typing.Dict[str, AbstractAction]) -> None
-        self.__on_error = kwargs.get('on_error')
-        self.__actions = set(actions)
+        # type: (typing.Tuple[AbstractAction, ...], typing.Dict[str, AbstractIdentifiedNode]) -> None
+        super(Parallel, self).__init__(identifier='auto-generated', on_error=kwargs.get('on_error'))
+        self.actions = set(copy.deepcopy(actions))
 
 
-class Kill(AbstractAction):
+class Switch(AbstractAction):
+    """Set of JSP Expression Language case statements and actions to follow as a result"""
+
+    def __init__(self, cases, default):
+        # type: Dict[str, AbstractIdentifiedNode], AbstractIdentifiedNode) -> None
+        super(Switch, self).__init__(identifier='auto-generated', on_error=None)
+        self.cases = set(copy.deepcopy(cases))
+        self.default = copy.deepcopy(default)
+
+
+class Kill(AbstractIdentifiedNode):
     """Stop execution immediately"""
 
     def __init__(self, message):
+        super(Kill, self).__init__(identifier='auto-generated')
         self.__message = message
--- a/pyoozie/workflow.py
+++ b/pyoozie/workflow.py
@@ -1,0 +1,43 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
+# Use of this source code is governed by a MIT-style license that can be found in the LICENSE file.
+import typing  # pylint: disable=unused-import
+
+
+class AbstractAction(object):
+    pass
+
+
+class Action(AbstractAction):
+    """Concrete action to execute (takes an action tag, e.g. pyoozie.Shell, pyoozie.Email. etc.)"""
+
+    def __init__(self, action_tag, on_error=None):
+        # type: (tags.ActionTag, AbstractAction) -> None
+        self.__action_tag = action_tag
+        self.__on_error = on_error
+
+
+class Sequence(AbstractAction):
+    """Set of actions to execute sequentially (implemented by chaining actions and 'OK' transitions)"""
+
+    def __init__(self, *actions, on_error=None):
+        # TODO type: (typing.Sequence[AbstractAction], AbstractAction) -> None
+        self.__on_error = on_error
+        self.__actions = actions
+
+
+class Parallelization(AbstractAction):
+    """Set of actions to execute in parallel (implemented as fork/join tag pair)"""
+
+    def __init__(self, *actions, on_error=None):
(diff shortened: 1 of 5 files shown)


tests/pyoozie/test_builder.py
```diff
@@ -130,82 +118,30 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
...
+                workflow.Sequence(
+                    workflow.Action(tags.Shell(exec_command='echo', arguments=("'C'",))),
+                    workflow.Action(tags.Shell(exec_command='echo', arguments=("'D'",))),
+                ),
+                on_error=workflow.Action(tags.Shell(exec_command='echo', arguments=("'A parallel action failed'",))),
+            ),
+            on_error=workflow.Kill('Flow failed')
+        ),
+    )
+    assert workflow_builder.build()
```
cfournie: If we use this interface then I think I'll remove the work `builder`; we're no longer using that pattern anymore.

pyoozie/workflow.py
```diff
@@ -0,0 +1,43 @@
...
+
+class AbstractAction(object):
+    pass
+
+
+class Action(AbstractAction):
+    """Concrete action to execute (takes an action tag, e.g. pyoozie.Shell, pyoozie.Email. etc.)"""
+
+    def __init__(self, action_tag, on_error=None):
+        # type: (tags.Shell, AbstractAction) -> None
```
cfournie: I'll let this accept more than just `tags.Shell` in the future.

pyoozie/builder.py
```diff
@@ -27,56 +27,15 @@ def _coordinator_submission_xml(username, coord_xml_path, configuration=None, in
...
-</workflow-app>
-'''.format(action_payload_xml=format_xml(self._action_payload.xml(indent=indent)),
-           action_error_xml=format_xml(self._action_error.xml(indent=indent)),
-           kill_message=self._kill_message,
-           action_name=self._action_name,
-           name=self._name).strip()
+    def __init__(self, name, parameters=None, configuration=None, credentials=None, job_tracker=None,
+                 name_node=None, job_xml_files=None, default_retry_max=None, default_retry_interval=None,
+                 default_retry_policy=None, actions=None):
+        # pylint: disable=unused-argument
```
honkfestival: This is fine for now, but what's your eventual plan with all of the parameters?

pyoozie/workflow.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2017 "Shopify inc." All rights reserved.
+# Use of this source code is governed by a MIT-style license that can be found in the LICENSE file.
+import typing  # pylint: disable=unused-import
+
+from pyoozie import tags  # pylint: disable=unused-import
+
+
+class AbstractAction(object):
```
honkfestival: What does it mean to be an `AbstractAction`? Scanning the subclasses below, it seems that the common element in all (except for `Kill`) is that they have an error handler. That makes sense, given what we talked about yesterday.

So, what do you think about
 - making `AbstractAction` an [ABC](https://docs.python.org/2/library/abc.html),
 - adding an abstract method `on_error()` to it, and
 - removing `Kill` from the hierarchy (since it doesn't make sense to have an error handler for `Kill`)?

Since this design is meant to make error handling tractable, I think it makes sense to be explicit about error handling intentions.

pyoozie/workflow.py
```diff
@@ -0,0 +1,43 @@
...
+class Sequence(AbstractAction):
+    """Set of actions to execute sequentially (implemented by chaining actions and 'OK' transitions)"""
+
+    def __init__(self, *actions, **kwargs):
+        # type: (typing.Tuple[AbstractAction, ...], typing.Dict[str, AbstractAction]) -> None
+        self.__on_error = kwargs.get('on_error')
+        self.__actions = actions
+
+
+class Parallelization(AbstractAction):
```
honkfestival: What do you think of renaming this one to `ForkJoin`?

pyoozie/workflow.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class AbstractAction(object):
+    pass
+
+
+class Action(AbstractAction):
+    """Concrete action to execute (takes an action tag, e.g. pyoozie.Shell, pyoozie.Email. etc.)"""
+
+    def __init__(self, action_tag, on_error=None):
```
honkfestival: I'm not sure if `None` is an appropriate default, but I'm also not sure what to use instead. Maybe some kind of `Defer`? How do you see error handlers composing? Or are we just going to walk up the graph until we find one?

tests/pyoozie/test_builder.py
```diff
@@ -130,82 +118,30 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
...
-    assert str(assertion_info.value) == 'Can only add one action in this version'
+@pytest.mark.xfail(raises=NotImplementedError)
+def test_workflow_builder():
+
+    workflow_builder = builder.WorkflowBuilder(
+        name='descriptive-name',
+        actions=workflow.Sequence(
+            workflow.Action(tags.Shell(exec_command='echo', arguments=("'A'",))),
+            workflow.Parallelization(
+                workflow.Sequence(
```
honkfestival: Does a `Parallelization` always take `Sequence`s or would bare `Action`s suffice? From the definition it looks like a bare `Action` is okay, just wondering what your intention is here.

pyoozie/workflow.py
```diff
@@ -0,0 +1,43 @@
...
+        self.__on_error = on_error
+
+
+class Sequence(AbstractAction):
+    """Set of actions to execute sequentially (implemented by chaining actions and 'OK' transitions)"""
+
+    def __init__(self, *actions, **kwargs):
+        # type: (typing.Tuple[AbstractAction, ...], typing.Dict[str, AbstractAction]) -> None
+        self.__on_error = kwargs.get('on_error')
+        self.__actions = actions
```
honkfestival: It's probably best to make a copy of the `actions` (and note it in the docstring) so that people can't change them later. Making a copy is also closer to the semantics of what's actually going to happen in the workflow.

pyoozie/workflow.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class Parallelization(AbstractAction):
+    """Set of actions to execute in parallel (implemented as fork/join tag pair)"""
+
+    def __init__(self, *actions, **kwargs):
+        # type: (typing.Tuple[AbstractAction, ...], typing.Dict[str, AbstractAction]) -> None
+        self.__on_error = kwargs.get('on_error')
+        self.__actions = set(actions)
```
honkfestival: Ditto here re: the copy, 👍  for the `set` though!

pyoozie/workflow.py
```diff
@@ -0,0 +1,43 @@
...
+    """Concrete action to execute (takes an action tag, e.g. pyoozie.Shell, pyoozie.Email. etc.)"""
+
+    def __init__(self, action_tag, on_error=None):
+        # type: (tags.Shell, AbstractAction) -> None
+        self.__action_tag = action_tag
+        self.__on_error = on_error
+
+
+class Sequence(AbstractAction):
+    """Set of actions to execute sequentially (implemented by chaining actions and 'OK' transitions)"""
```
honkfestival: nit: `Sequence of actions...`

pyoozie/workflow.py
kmtaylor-github: Again, the `set()` should result in a copy.

tests/pyoozie/test_builder.py
kmtaylor-github: In my mental model all three should be equivalent -- a bare `Action` is a degenerate `Parallelization` or `Serialization`

tests/pyoozie/test_builder.py
kmtaylor-github: Ditto for `build`, I suspect.

pyoozie/workflow.py
cfournie: > making AbstractAction an ABC,

Definitely

> adding an abstract method on_error() to it, and

👍 

> removing Kill from the hierarchy (since it doesn't make sense to have an error handler for Kill)?

Yeah `Kill` shouldn't be in the hierarchy, doesn't have error handling (it causes execution, parallel or not, to stop). The only concrete actions that we would support to start are `Shell`, `Email`, and `SubWorkflow` ([others can be found in various XML schemas](https://github.com/apache/oozie/blob/master/client/src/main/resources/oozie-workflow-0.5.xsd#L148)).

pyoozie/workflow.py
cfournie: > Or are we just going to walk up the graph until we find one?

I was planning on that. Much like raising an exception, go up the stack until there's a handler execute it, and maybe even continue raising.

pyoozie/workflow.py
cfournie: > Isn't the vargs capture implicitly a new list?

Even better, it's a new tuple 😁

pyoozie/workflow.py
cfournie: > What do you think of renaming this one to `ForkJoin`?

To be orthogonally consistent, `Sequence` should also be renamed to using Oozie terminology, e.g. `OnOk`. Do we break with Oozie terminology altogether with these structures or try to shoehorn Oozie terms in?

`ForkJoin` and OnOk` v.s. `Serial`/`Parallel` v.s. `Sequence`/`Parallelization`; which is best? Or is there a third?

pyoozie/workflow.py
cfournie: Set copies the collection but not the items; did you want me to perform a deep copy? I use set here (and will document so) because I want to discourage any reliance upon action order here.

tests/pyoozie/test_builder.py
cfournie: > In my mental model all three should be equivalent

Correct.

`Parallelization` should take either another `Parallelization` (i.e. nested fork/join), a concrete `Action` (`Shell`, `Email`, `SubWorkflow`), or `Sequence`.
`Serial` should also accept another `Parallelization` or `Sequence` (allowing a sequence to share the same error handler).

tests/pyoozie/test_builder.py
cfournie: Yes; dropping this in favour of `.xml()` calles on `pyoozie.tags` objects (see https://github.com/Shopify/pyoozie/pull/27).

pyoozie/builder.py
cfournie: Going to get rid of this class and build [something more like this](https://github.com/Shopify/pyoozie/pull/27/files#diff-ccc9805975972566242e305d3d35aa79R344).

pyoozie/workflow.py
honkfestival: ack sorry, I meant a `deepcopy`. It's not the collection I'm worried about (because of the splat); it's the instances themselves.

pyoozie/workflow.py
honkfestival: I have a mild preference for `Parallel` vs `Parallelization`, so a mild vote goes to `Serial`/`Parallel`. I think your point about orthogonal consistency is more important though. 👍

tests/pyoozie/test_builder.py
honkfestival: So the `Sequence` with only a single element isn't tended as advice, just an illustration?

cfournie: Just an illustration, yes.

pyoozie/workflow.py
honkfestival: Yes, another `deepcopy`.

PR #29, opened 2017-03-27
Add types to pyoozie.tags

cfournie:
This PR beings to add type hinting to pyoozie by:
  - Adding a `make type` target to speed up testing; and
  - Adds type annotations to `pyoozie.tags`.

During that, a return type change was made: XML output is returned as UTF-8 encoded bytes and not unicode strings. `pyoozie.client.jobs_submit_coordinator` and `pyoozie.client.jobs_submit_workflow`, which both XML output in HTTP POST bodies to Oozie were tested against Oozie 4.1.0 to ensure that they still submit data correctly (which uncovered an issue reading non-ASCII coordinator configurations shown in https://github.com/Shopify/pyoozie/pull/31).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,13 +27,17 @@ autopep8:
 		@echo 'Auto Formatting...'
 		@$(python_files) | xargs -0 autopep8 --max-line-length 120 --jobs 0 --in-place --aggressive
 
-lint:
-		@echo 'Linting...'
-		@pylint --rcfile=pylintrc setup.py pyoozie tests
+type:
 		@if [ "$(python_version_major)" = "3" ]; then \
 			echo 'Checking type annotations...'; \
 			mypy --py2 pyoozie tests --ignore-missing-imports; \
 		fi
+
+sourcelint:
+		@echo 'Linting...'
+		@pylint --rcfile=pylintrc setup.py pyoozie tests
 		@pep8
+
+lint: sourcelint type
 
 autolint: autopep8 lint
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -19,8 +19,8 @@ COMPILED_REGEX_IDENTIFIER = re.compile(REGEX_IDENTIFIER)
 ALLOWABLE_NAME_CHARS = set(string.ascii_letters + string.punctuation + string.digits + ' ')
 ONE_HUNDRED_YEARS = 100 * 365.24
 
-PropertyValuesType = typing.Optional[typing.Dict[typing.Text, typing.Any]]
-JobXmlFilesType = typing.Optional[typing.Iterable[typing.Text]]
+PropertyValuesType = typing.Dict[typing.Text, typing.Any]
+JobXmlFilesType = typing.Iterable[typing.Text]
 
 
 class ExecutionOrder(enum.Enum):
@@ -116,7 +116,7 @@ class _PropertyList(XMLSerializable, dict):
             self,
             xml_tag,          # type: typing.Text
             attributes=None,  # type: typing.Optional[typing.Dict[typing.Text, typing.Text]]
-            values=None       # type: PropertyValuesType
+            values=None       # type: typing.Optional[PropertyValuesType]
     ):
         # type: (...) -> None
         super(_PropertyList, self).__init__(xml_tag=xml_tag)
@@ -147,7 +147,7 @@ class Parameters(_PropertyList):
     """
 
     def __init__(self, values=None):
-        # type: (PropertyValuesType) -> None
+        # type: (typing.Optional[PropertyValuesType]) -> None
         super(Parameters, self).__init__(xml_tag='parameters', values=values)
 
 
@@ -155,7 +155,7 @@ class Configuration(_PropertyList):
     """Coordinator job submission, workflow, workflow action configuration XML."""
 
     def __init__(self, values=None):
-        # type: (PropertyValuesType) -> None
+        # type: (typing.Optional[PropertyValuesType]) -> None
         super(Configuration, self).__init__(xml_tag='configuration', values=values)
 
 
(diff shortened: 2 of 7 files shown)


pyoozie/tags.py
```diff
@@ -69,19 +74,21 @@ class XMLSerializable(object):
     __metaclass__ = abc.ABCMeta
 
     def __init__(self, xml_tag):
+        # type: (typing.Text) -> None
         self.xml_tag = xml_tag
 
     def xml(self, indent=False):
+        # type: (bool) -> str
```
cfournie: Since we define the doc as being utf8, let's return a byte string and explicitly encode as utf-8.

pyoozie/tags.py
```diff
@@ -174,22 +188,36 @@ def __init__(self, values, credential_name, credential_type):
...
+            capture_output=False  # type: bool
+    ):
+        # type: (...) -> None
         super(Shell, self).__init__(xml_tag='shell')
         self.exec_command = exec_command
         self.job_tracker = job_tracker
         self.name_node = name_node
-        self.prepares = prepares if prepares else list()
-        self.job_xml_files = job_xml_files if job_xml_files else list()
+        self.prepare = prepare if prepare else []
```
cfournie: Tag name is singular, so let's rename the parameter accordingly.

tests/pyoozie/test_xml.py
```diff
@@ -35,7 +36,7 @@ def workflow_builder():
         name='payload',
         action=tags.Shell(exec_command='echo "test"'),
         action_on_error=tags.Email(to='person@example.com', subject='Error', body='A bad thing happened'),
-        kill_on_error='Failure message',
+        kill_on_error='Failure message 😢',
```
cfournie: Let's make sure that our utf-8 encoding works

Makefile
```diff
@@ -27,13 +27,17 @@ autopep8:
...
-lint:
-		@echo 'Linting...'
-		@pylint --rcfile=pylintrc setup.py pyoozie tests
+type:
 		@if [ "$(python_version_major)" = "3" ]; then \
 			echo 'Checking type annotations...'; \
 			mypy --py2 pyoozie tests --ignore-missing-imports; \
 		fi
+
+sourcelint:
```
JasonMWhite: Why `lint` -> `sourcelint` ?

JasonMWhite: Ah, the new make task below.

pyoozie/tags.py
```diff
@@ -299,7 +342,17 @@ def _xml(self, doc, tag, text):
...
     """Email action for use within a workflow."""
 
-    def __init__(self, to, subject, body, cc=None, bcc=None, content_type=None, attachments=None):
+    def __init__(
+            self,
+            to,  # type: typing.Union[typing.Text, typing.Iterable[typing.Text]]
+            subject,  # type: typing.Text
+            body,  # type: typing.Text
+            cc=None,  # type: typing.Union[typing.Text, typing.Iterable[typing.Text]]
+            bcc=None,  # type: typing.Union[typing.Text, typing.Iterable[typing.Text]]
```
JasonMWhite: `None` is neither of the allowable types here. Missing a `typing.Optional` ?

pyoozie/tags.py
```diff
@@ -310,11 +363,13 @@ def __init__(self, to, subject, body, cc=None, bcc=None, content_type=None, atta
         self.attachments = attachments
 
     def _xml(self, doc, tag, text):
-        def format_list(emails):
-            if hasattr(emails, '__iter__') and not isinstance(emails, str):
-                return ','.join(sorted(emails))
+        # type: (yattag.doc.Doc, typing.Callable, typing.Callable) -> yattag.doc.Doc
```
JasonMWhite: More specific about the types of `types.Callable`?

pyoozie/tags.py
```diff
@@ -341,9 +396,22 @@ def format_list(emails):
...
+            frequency,  # type: int
+            start,  # type: datetime.datetime
+            end=None,  # type: datetime.datetime
+            timezone=None,  # type: typing.Text
+            workflow_configuration=None,  # type: typing.Dict[typing.Text, typing.Text]
+            timeout=None,  # type: int
+            concurrency=None,  # type: int
+            execution_order=None,  # type: ExecutionOrder
+            throttle=None,  # type: int
+            parameters=None  # type: typing.Dict[typing.Text, typing.Text]
```
JasonMWhite: Many more `typing.Optional`s here

pyoozie/tags.py
```diff
@@ -16,6 +19,9 @@
 ALLOWABLE_NAME_CHARS = set(string.ascii_letters + string.punctuation + string.digits + ' ')
 ONE_HUNDRED_YEARS = 100 * 365.24
 
+PropertyValuesType = typing.Optional[typing.Dict[typing.Text, typing.Any]]
```
honkfestival: Where does `typing.Text` come from? It's not in the docs for Python 3.5.2, according to Dash.

honkfestival: Also, what do you think about leaving the `typing.Optional` off of this type declaration? That way you can choose whether or not to accept `None` on a per-usage basis.

(I personally see optionality as being dictated by usage rather than being an essential part of a type.)

JasonMWhite: I had the same idea, `typing.Optional[PropertyValuesType]` seems more natural here.

cfournie: I'm not sure which iteration of Python 3.5 that it was added in, but it's in [the latest Python 3.5 docs](https://docs.python.org/3.5/library/typing.html#typing.Text). For other Python versions it comes from the [PyPi `typing` package](https://pypi.python.org/pypi/typing).

> off of this type declaration? 

Good idea 👍

PR #35, opened 2017-03-28
Add workflow tag

cfournie:
This PR adds a basic `workflow-app` tag that simply contains a `start` and `end` with no other work and validates the XML that is generated from it.

An alias is not added to the `pyoozie` package and a warning is issued if it is used because it is not yet finished.

Subsequent PRs will add functionality discussed in https://github.com/Shopify/pyoozie/pull/26, including:
  1. APIs to add a single action;
  2. APIs to add an action on error (allowing us to remove `WorkflowBuilder`);
  3. APIs to add multiple serial action (with actions on error); and then
  4. APIs to add multiple parallel actions (with actions on error).

### DIFF
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -42,20 +42,6 @@ EXEC_LIFO = ExecutionOrder.LIFO
 EXEC_NONE = ExecutionOrder.NONE
 
 
-class RetryPolicy(enum.Enum):
-    """Retry policy for workflow actions"""
-
-    EXPONENTIAL = 'EXPONENTIAL'
-    PERIODIC = 'PERIODIC'
-
-    def __str__(self):
-        return self.value
-
-
-RETRY_EXPONENTIAL = RetryPolicy.EXPONENTIAL
-RETRY_PERIODIC = RetryPolicy.PERIODIC
-
-
 def validate_xml_name(name):
     # type: (typing.Text) -> typing.Text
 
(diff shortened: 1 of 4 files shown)


pyoozie/tags.py
```diff
@@ -491,3 +506,55 @@ def _xml(self, doc, tag, text):
...
+            job_tracker=job_tracker,
+            name_node=name_node,
+            job_xml_files=job_xml_files,
+            configuration=configuration
+        )
+        self.credentials = credentials
+
+        self.default_retry_max = default_retry_max
+        self.default_retry_interval = default_retry_interval
+        self.default_retry_policy = default_retry_policy
```
kmtaylor-github: Unused?
Should we validate these? (e.g. >=0)

cfournie: Ah, we shouldn't even include them here probably; this is copy-pasta from a previous incarnation of this API.

pyoozie/tags.py
```diff
@@ -491,3 +492,48 @@ def _xml(self, doc, tag, text):
...
+    def __init__(
+            self,
+            name,                         # type: typing.Text
+            parameters=None,              # type: typing.Optional[PropertyValuesType]
+            configuration=None,           # type: typing.Optional[PropertyValuesType]
+            credentials=None,             # type: typing.Optional[typing.Iterable[Credential]]
+            job_tracker=None,             # type: typing.Optional[typing.Text]
+            name_node=None,               # type: typing.Optional[typing.Text]
+            job_xml_files=None,           # type: typing.Optional[JobXmlFilesType]
+    ):
```
honkfestival: :(

PR #37, opened 2017-03-29
Remove dependency links

cfournie:
This is a [deprecated feature in pip](https://github.com/pypa/pip/blame/master/NEWS.rst#L690), so let's not rely upon it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ clean:
 		rm -fr *.egg-info
 
 install:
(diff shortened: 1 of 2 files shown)


erikwright:
May I suggest loosening the version constraint somewhat? According to your versioning strategy, >= 0.2.0 and < 1 should be compatible.

erikwright:
LGTM

PR #39, opened 2017-03-30
Replace pep8 with pycodestyle

cfournie:
[pep8 was renamed](http://pep8.readthedocs.io/en/release-1.7.x/) [pycodestyle](https://github.com/PyCQA/pycodestyle).


angelini:
Do we have to manually edit all projects using `shopify_python` when we update it?

JasonMWhite:
Technically, it's unrelated from using `shopify_python` right now. They each independently use it.

PR #40, opened 2017-03-30
Define dependency versions

cfournie:
This PR defines minimum versions for `install_requires` and pins `extras_require` to exact versions.


setup.py
```diff
@@ -28,30 +28,27 @@ def get_version():
...
+        'six >= 1.10.0',
+        'typing >= 3.6.1',
+        'untangle >= 1.1.0',
+        'yattag >= 1.7.2',
+
     ],
     extras_require={
-        'deploy': [
-            'setuptools>=0.9'
-        ],
```
cfournie: We don't strictly need this

JasonMWhite:
LGTM, but @erikwright has been looking into this a lot more than me recently.

setup.py
```diff
@@ -28,30 +28,26 @@ def get_version():
...
+            'mypy == 0.501; python_version >= "3.3"',
             'pycodestyle == 2.2.0',
-            'pylint',
-            'pytest-cov',
-            'pytest-randomly',
-            'pytest>=2.7',
-            'requests-mock',
-            'shopify_python==0.2.2',
-            'xmltodict',
+            'pylint == 1.6.5',
```
erikwright: Why be more specific for `test` than otherwise?

cfournie: To encourage reproducible builds, I set `test` requirements to exact versions. I'd do the same for `install_requires` requirements but we want to allow library users some version flexibility to avoid dependency conflicts.

erikwright:
LGTM.

setup.py
erikwright: Sorry, thought about it some more and I'm not sure this makes sense. You are explicitly setting the versions of your immediate dependencies, but will you also list all of your transitive dependencies? If not, you are still susceptible to non-deterministic builds.

If you want really deterministic builds I think you need to actually install the packages, let Pip pick whatever it does, and then freeze it to a requirements file. I would argue that it's not a requirement. In fact, you might as well have something that will fail (and tell you) if/when a new version of a dependency is released (that meets your production requirements list) that doesn't pass your tests.

Issue #41, opened 2017-04-07
Sparse Coordinator Objects


honkfestival:
**TL;DR** - please consider separate result type for the Jobs API and the Job API

It's a little confusing right now for the Jobs and Job APIs to reuse the same objects (e.g. `pyoozie.model.Coordinator`).

In particular, I didn't realise that an empty `conf` property meant I needed to use `get_coordinator_info` from the Job API to have the `conf` property be populated.

I understand that the Oozie APIs work this way, but having separate types would make it easier to discover that you need a separate call to get all of the job details.

Mild suggestions:

 - `CoordinatorReference` and `Coordinator`
 - `Coordinator` and `CoordinatorInfo`
 - `Coordinator` and `CoordinatorDetails`

These might make the most sense in a porcelain-style section of the API with the plumbing still left for experts.

@cfournie

kmtaylor-github:
What call did you make that retrieved a coordinator with a null `conf` object? Most (all?) APIs should populate the details.

This whole scenario is a huge pain point. For example, the janky script I wrote for DD144 to pause all coordinators took ~15mins to run due to the extra overhead of fetching `conf` details for each of our ~700 coordinators; bypassing that would cut the time significantly.

The problem with the "different object" approach is that there is no clarity in the Oozie API about what is and isn't retrieved on any particular call. For example, (some|all|none) of the actions will be present based on the queries you did.

An alternate approach would be to lazily update these fields on access. We'd need a magic marker to distinguish between "dunno" and "nope".

------

On a somewhat related note, the current API approach for the JobAPI is very focused around ID/name, which means that many usecases will query coord/wf info repeatedly. Some API allow an optional `coordinator` kwarg and use that preferentially.

honkfestival:
I was using `jobs_all_active_coordinators`, which currently returns `{}` for the `conf` object.

> The problem with the "different object" approach is that there is no clarity in the Oozie API about what is and isn't retrieved on any particular call.

Yeah, there are really two problems here:
1. I couldn't tell if it was actually `{}` or if I was doing it wrong.
1. Upon learning that I _was_ doing it wrong, I then had to make `n` network calls to get the info I did want.

Lazy fields would have solved the first problem, but not the second. Unfortunately, I think it's on Oozie itself to solve the second one.

("pull requests welcome")

kmtaylor-github:
> I was using jobs_all_active_coordinators, which currently returns {} for the conf object.

[...]

PR #42, opened 2017-04-18
Add ability to update a coordinator's definition

sabidib:
This adds the ability to update a coordinator's definition without killing and recreating the coordinator.

According to the [docs]( https://oozie.apache.org/docs/4.1.0/DG_CommandLineTool.html#Updating_coordinator_definition_and_properties) a coordinator's definition can only be updated if the following are not changed:
  - coordinator name
  - frequency 
  - start time
  - end time
  - timezone

If any of these are changed, then there should be a hard failure on the [request](https://github.com/Shopify/pyoozie/blob/master/pyoozie/client.py#L144).

I couldn't find anything specific on what coordinator states allow for an update so I assumed we would only want to update coordinators that are `active` or `running` or `suspended`. 

Curious if updating a coordinator in a `suspended` state makes sense.
@kmtaylor-github @cfournie 
Thoughts?

### DIFF
--- a/pyoozie/client.py
+++ b/pyoozie/client.py
@@ -570,7 +570,7 @@ class OozieClient(object):
                 self.logger.info('Preparing to update coordinator %s:\n%s', xml_path, conf)
             reply = self._put('job/{}?action=update'.format(coordinator_id), conf)
 
-            if 'update' not in reply:
+            if not reply or 'update' not in reply:
                 raise exceptions.OozieException.operation_failed('update coordinator')
 
             if self._verbose:
--- a/pyoozie/client.py
+++ b/pyoozie/client.py
@@ -146,8 +146,8 @@ class OozieClient(object):
     def _get(self, endpoint, content_type=None):
         return self._request('GET', endpoint, content_type)
 
-    def _put(self, endpoint, content_type='application/xml'):
-        return self._request('PUT', endpoint, content_type)
+    def _put(self, endpoint, content=None, content_type='application/xml'):
+        return self._request('PUT', endpoint, content_type, content)
 
     def _post(self, endpoint, content, content_type='application/xml'):
         return self._request('POST', endpoint, content_type, content)
@@ -545,7 +545,7 @@ class OozieClient(object):
         return False
 
     # ===========================================================================
-    # Job API - submit jobs
+    # Job API - submit and update jobs
     # ===========================================================================
 
     def jobs_submit_coordinator(self, xml_path, configuration=None):
@@ -560,6 +560,25 @@ class OozieClient(object):
             coord = self.job_coordinator_info(coordinator_id=reply['id'])
             return coord
         raise exceptions.OozieException.operation_failed('submit coordinator')
+
+    def jobs_update_coordinator(self, coordinator_id, xml_path, configuration=None):
+        user = self._user or 'oozie'
+        coord = self._fetch_coordinator_or_action(coordinator_id)
+        if coord.status.is_active():
+            conf = xml._coordinator_submission_xml(user, xml_path, configuration=configuration)
+            if self._verbose:
+                self.logger.info('Preparing to update coordinator %s:\n%s', xml_path, conf)
+            reply = self._put('job/{}?action=update'.format(coordinator_id), conf)
+
+            if 'update' not in reply:
+                raise exceptions.OozieException.operation_failed('update coordinator')
+
+            if self._verbose:
+                self.logger.info('Coordinator %s updated with diff %s', coordinator_id, reply['update']['diff'])
+
+            return self.job_coordinator_info(coordinator_id=coordinator_id)
+        else:
+            raise exceptions.OozieException.operation_failed('coordinator status must be active in order to update')
 
     def jobs_submit_workflow(self, xml_path, configuration=None, start=False):
         user = self._user or 'oozie'
--- a/tests/pyoozie/test_client.py
+++ b/tests/pyoozie/test_client.py
@@ -1238,6 +1238,7 @@ class TestOozieClientJobSubmit(object):
         with mock.patch.object(api, '_post') as mock_post:
             with mock.patch.object(api, 'job_coordinator_info') as mock_info:
                 mock_info.return_value = sample_coordinator_running
+
                 mock_post.return_value = None
                 with pytest.raises(exceptions.OozieException) as err:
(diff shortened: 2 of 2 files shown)


kmtaylor-github:
> Curious if updating a coordinator in a suspended state makes sense.

Yes! This is actually a really important usecase. A deploy shouldn't skip a suspended coord, nor should it resume it.

pyoozie/client.py
```diff
@@ -505,6 +505,13 @@ def job_coordinator_resume(self, coordinator_id=None, name=None, user=None):
             return True
         return False
 
+    def job_coordinator_update(self, coordinator_id=None, name=None, user=None):
+        coord = self._fetch_coordinator_or_action(coordinator_id, name, user)
+        if coord.status.is_active() or coord.status.is_running() or coord.status.is_suspended():
```
kmtaylor-github: Isn't this highly redundant? `RUNNING` and `SUSPENDED` should both be covered by `is_active`.

pyoozie/client.py
```diff
@@ -505,6 +505,13 @@ def job_coordinator_resume(self, coordinator_id=None, name=None, user=None):
...
+            self._coordinator_perform_simple_action(coord, 'update')
```
kmtaylor-github: I don't think a simple action will suffice here. The update accepts new parameters from the XML payload of the request.

pyoozie/client.py
```diff
@@ -561,6 +561,25 @@ def jobs_submit_coordinator(self, xml_path, configuration=None):
...
+    def jobs_update_coordinator(self, xml_path, coordinator_id, configuration=None):
+        user = self._user or 'oozie'
+        coord = self._fetch_coordinator_or_action(coordinator_id)
+        if coord.status.is_active():
+            conf = xml._coordinator_submission_xml(user, xml_path, configuration=configuration)
+            if self._verbose:
+                self.logger.info('Preparing to update coordinator %s:\n%s', xml_path, conf)
+            reply = self._put('job/{}?action=update'.format(coordinator_id), conf)
+
+            if 'update' not in reply:
```
kmtaylor-github: Should probably nullcheck reply too, just to be safe

pyoozie/client.py
```diff
@@ -561,6 +561,25 @@ def jobs_submit_coordinator(self, xml_path, configuration=None):
...
+                raise exceptions.OozieException.operation_failed('update coordinator')
+
+            if self._verbose:
+                self.logger.info('Coordinator %s updated with diff %s', coordinator_id, reply['update']['diff'])
+
+            return self.job_coordinator_info(coordinator_id=coordinator_id)
```
kmtaylor-github: I'm a bit torn. This is an obvious return value, but the diffs seem relevant too.

sabidib:
I gave this a test spin with a local instance of oozie and switched a coordinator between the [minimal coordinator](https://github.com/Shopify/pyoozie/blob/master/tests/data/minimal_coordinator.xml) and the [full coordinator](https://github.com/Shopify/pyoozie/blob/master/tests/data/full_coordinator.xml) successfully. The diffs made sense as well. It also handled "updating" to the same coordinator xml showing an empty diff.

kmtaylor-github:
>  It also handled "updating" to the same coordinator xml showing an empty diff.

Presumably this was with identical config data?

sabidib:
> Presumably this was with identical config data?
Yes, the configuration was the same.

pyoozie/client.py
sabidib: I agree, my idea is that you could extract the "diff" by looking at the coordinator result and comparing it to the input coordinator. Returning the diff result string doesn't seem to fit the rest of the API.

pyoozie/client.py
```diff
@@ -561,6 +561,25 @@ def jobs_submit_coordinator(self, xml_path, configuration=None):
...
+    def jobs_update_coordinator(self, coordinator_id, xml_path, configuration=None):
+        user = self._user or 'oozie'
+        coord = self._fetch_coordinator_or_action(coordinator_id)
+        if coord.status.is_active():
+            conf = xml._coordinator_submission_xml(user, xml_path, configuration=configuration)
+            if self._verbose:
+                self.logger.info('Preparing to update coordinator %s:\n%s', xml_path, conf)
+            reply = self._put('job/{}?action=update'.format(coordinator_id), conf)
+
+            if not reply and 'update' not in reply:
```
cfournie: Should this be `or`?

pyoozie/client.py
cfournie: What about a subclass of `Coordinator` called `CoordinatorUpdate` that has a diff attribute?

pyoozie/client.py
kmtaylor-github: What does the reply look when the op fails? In particular, the case where an un-update-able field is changed may be worth handling with a different exception.

pyoozie/client.py
sabidib: I like that idea ^

pyoozie/client.py
sabidib: It should, good catch

sabidib: The request should throw an exception. Let me get the exact error message to see how helpful it is

pyoozie/client.py
kmtaylor-github: Big 👎  to the subclass strategy from me (related: https://github.com/Shopify/pyoozie/issues/41)

Do we want a `CoordinatorUpdateWithDetails`?  `CoordinatorDetailsWithoutUpdate`?

sabidib: How about something like `CoordinatorUpdateDiff` with a reference to the input `Coordinator` object, the new `Coordinator` object, and the oozie diff response?

pyoozie/client.py
sabidib: If I try to update the coordinator with a new one that has a different frequency, the request throws a nice error:

If I changed the frequency the request throws an exception...That's actually helpful:
```
OozieCommunicationException: E0803: IO error, E1023: Coord Job update Error: [Frequency can't be changed. Old frequency = 140 new frequency = 1440]
```

pyoozie/client.py
kmtaylor-github: Talked IRL: the diff really doesn't matter that much. It gets logged in verbose mode, that's enough for me.

sabidib:
Any outstanding comments?

PR #44, opened 2017-04-27
Add missing is_active tags to some statuses

kmtaylor-github:
The `READY`/`WAITING` statuses for coordinator actions weren't tagged as being "active". 
Ditto for `PREP` on workflow actions.

@cfournie @sabidib


sabidib:
👍  LGTM

PR #45, opened 2017-05-01
Add test utilities

cfournie:
This PR adds/changes utilities to make testing Oozie workflow XML a bit easier.  Specifically:
- `ParsedXml` will parse an XML string once using `ElementTree` and then allows users to assert using XPath that a single tag:
  - Exists; or
  - Has some specific attribute values
- `assert_valid_workflow` checks to make sure that duplicate names are not defined (and fixes an encoding issue)

These are used to test XML workflow generation in https://github.com/Shopify/pyoozie/pull/46.

### DIFF
--- a/tests/pyoozie/test_xml.py
+++ b/tests/pyoozie/test_xml.py
@@ -43,7 +43,7 @@ def test_workflow_submission_xml(username, workflow_app_path):
         workflow_xml_path=workflow_app_path,
         indent=True,
     )
-    assert tests.utils.xml_to_dict_unordered('''
+    assert tests.utils.xml_to_comparable_dict('''
     <configuration>
         <property>
             <name>oozie.wf.application.path</name>
@@ -53,7 +53,7 @@ def test_workflow_submission_xml(username, workflow_app_path):
             <name>user.name</name>
             <value>test</value>
         </property>
-    </configuration>''') == tests.utils.xml_to_dict_unordered(actual)
+    </configuration>''') == tests.utils.xml_to_comparable_dict(actual)
 
 
 def test_workflow_submission_xml_with_configuration(username, workflow_app_path):
@@ -66,7 +66,7 @@ def test_workflow_submission_xml_with_configuration(username, workflow_app_path)
         indent=True
     )
 
-    assert tests.utils.xml_to_dict_unordered('''
+    assert tests.utils.xml_to_comparable_dict('''
     <configuration>
         <property>
             <name>oozie.wf.application.path</name>
@@ -80,7 +80,7 @@ def test_workflow_submission_xml_with_configuration(username, workflow_app_path)
             <name>user.name</name>
             <value>test</value>
         </property>
-    </configuration>''') == tests.utils.xml_to_dict_unordered(actual)
+    </configuration>''') == tests.utils.xml_to_comparable_dict(actual)
 
 
 def test_coordinator_submission_xml(username, coord_app_path):
@@ -89,7 +89,7 @@ def test_coordinator_submission_xml(username, coord_app_path):
         coord_xml_path=coord_app_path,
         indent=True
     )
-    assert tests.utils.xml_to_dict_unordered('''
+    assert tests.utils.xml_to_comparable_dict('''
     <configuration>
         <property>
             <name>oozie.coord.application.path</name>
@@ -99,7 +99,7 @@ def test_coordinator_submission_xml(username, coord_app_path):
             <name>user.name</name>
             <value>test</value>
         </property>
-    </configuration>''') == tests.utils.xml_to_dict_unordered(actual)
+    </configuration>''') == tests.utils.xml_to_comparable_dict(actual)
 
 
 def test_coordinator_submission_xml_with_configuration(username, coord_app_path):
@@ -111,7 +111,7 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
         },
         indent=True
     )
-    assert tests.utils.xml_to_dict_unordered('''
+    assert tests.utils.xml_to_comparable_dict('''
     <configuration>
         <property>
             <name>oozie.coord.application.path</name>
@@ -125,7 +125,7 @@ def test_coordinator_submission_xml_with_configuration(username, coord_app_path)
             <name>user.name</name>
             <value>test</value>
         </property>
-    </configuration>''') == tests.utils.xml_to_dict_unordered(actual)
+    </configuration>''') == tests.utils.xml_to_comparable_dict(actual)
 
 
 def test_workflow_builder(workflow_builder):
@@ -134,7 +134,7 @@ def test_workflow_builder(workflow_builder):
 
     # Is this XML expected
     actual_xml = workflow_builder.build()
-    assert tests.utils.xml_to_dict_unordered(expected_xml) == tests.utils.xml_to_dict_unordered(actual_xml)
+    assert tests.utils.xml_to_comparable_dict(expected_xml) == tests.utils.xml_to_comparable_dict(actual_xml)
 
     # Does it validate against the workflow XML schema?
     tests.utils.assert_valid_workflow(actual_xml)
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    document_dict = tests.utils.xml_to_dict_unordered("""
+<root>
+  <tag />
+  <tag key="value" />
+  <tag>Text</tag>
+</root>
+    """.strip())
+    assert document_dict == {
+        'root': {
+            'tag': [None, u'Text', {'@key': 'value'}]
```
sabidib: What does `<tag key="value">Text</tag>` look like? Does that throw an exception if it was included here?
What does `<tag different-key="different-value> Text </tag>` look like?

tests/test_utils.py
```diff
@@ -0,0 +1,124 @@
...
+<workflow-app xmlns="uri:oozie:workflow:0.5" name="descriptive-name">
+    <action name="alpha"/>
+    <action name="alpha"/>
+    <action name="beta"/>
+</workflow-app>""".strip()
+        assert len(et.fromstring(xml)), 'Invalid test XML'
+        tests.utils.assert_valid_workflow(xml)
+    assert str(assertion_info.value) == 'Name(s) reused: alpha'
+
+    # With valid XML, a missing a start tag should result in a schema violation assertion error
```
sabidib: Remove the extra `a` here.

tests/test_utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Asserting that a node that exists (when it does multiple times) should raise an assertion error
+    with pytest.raises(AssertionError) as assertion_info:
+        app.assert_node('/parameters/property')
+    assert str(assertion_info.value) == str("Found more than one resolution of xpath '/parameters/property'")
+
+    # Asserting that a node has a specific attribute should pass
+    app.assert_node('/start', to='end')
+
+    # Asserting that a node has a specific attribute value that it doens't should raise an assertion error
```
sabidib: `doesn't`

tests/test_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    with pytest.raises(AssertionError) as assertion_info:
+        app.assert_node('/start', to=str('not_end'))
```
sabidib: Why is there a `str` call here?

tests/test_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    assert str(assertion_info.value) == str("{'to': 'not_end'} != {'to': 'end'}")
+
+    # Asserting that a node has a specific attribute that it's missing should raise an assertion error
+    with pytest.raises(AssertionError) as assertion_info:
+        app.assert_node('/start', not_a_key=str('end'))
```
sabidib: Also `str` here?

tests/utils.py
```diff
@@ -19,7 +23,45 @@ def unorder(value):
     return unorder(xmltodict.parse(xml))
 
 
+NAMESPACE_ATTRIBUTE = re.compile(' xmlns="[^"]+"', flags=re.UNICODE)
+
+
+class ParsedXml(object):
```
sabidib: 👍  maybe something indicating this is test related?

tests/utils.py
```diff
@@ -19,7 +23,45 @@ def unorder(value):
     return unorder(xmltodict.parse(xml))
 
 
+NAMESPACE_ATTRIBUTE = re.compile(' xmlns="[^"]+"', flags=re.UNICODE)
```
prabcs: As oozie workflow xml [supports](https://issues.apache.org/jira/browse/OOZIE-2400) xml namespaces with prefix, this regex doesn't handle that -

```python
In [9]: patt = re.compile(' xmlns="[^"]+"', flags=re.UNICODE)

In [10]: xml_string = """
    ...: <workflow-app xmlns:h="uri:oozie:workflow:0.5" name="descriptive-name">
    ...:     <parameters>
    ...:         <property>
    ...:             <name>property_key</name>
    ...:             <value>property_value</value>
    ...:         </property>
    ...:         <property>
    ...:             <name>other_property_key</name>
    ...:             <value>🤣</value>
    ...:         </property>
    ...:     </parameters>
    ...:     <global>
    ...:         <job-tracker>job-tracker</job-tracker>
    ...:         <name-node>name-node</name-node>
    ...:     </global>
    ...:     <start to="end" />
    ...:     <end name="end" />
    ...: </workflow-app>
    ...: """.strip()

In [11]: patt.sub(str(''), xml_string.strip())
Out[11]: '<workflow-app xmlns:h="uri:oozie:workflow:0.5" name="descriptive-name">\n    <parameters>\n        <property>\n            <name>property_key</name>\n            <value>property_value</value>\n        </property>\n        <property>\n            <name>other_property_key</name>\n            <value>\xf0\x9f\xa4\xa3</value>\n        </property>\n    </parameters>\n    <global>\n        <job-tracker>job-tracker</job-tracker>\n        <name-node>name-node</name-node>\n    </global>\n    <start to="end" />\n    <end name="end" />\n</workflow-app>'

```

Not sure how it affects this PR though

prabcs: Also, there is some hypothetical case where the xml attribute can be something like - 
```xml
...
<tag xmlns="..."
...
```
 Should we worry about it ?

tests/utils.py
```diff
@@ -19,7 +23,45 @@ def unorder(value):
...
+
+
+class ParsedXml(object):
+
+    def __init__(self, xml_string):
+        if isinstance(xml_string, six.binary_type):
```
prabcs: Maybe document somewhere that any binary data in the xml string will be decoded as 'utf-8'.

tests/utils.py
```diff
@@ -34,7 +76,7 @@ def assert_valid_workflow(xml):
                 shell=True
             )
         except subprocess.CalledProcessError as e:
-            raise AssertionError('An XML validation error\n\n{error}\n\noccurred while parsing:\n\n{xml}'.format(
+            raise AssertionError('An XML validation error occurred.\n\n{error}\n\nParsing:\n\n{xml}'.format(
                 error=e.output.decode('utf8').strip(),
-                xml=xml,
+                xml=xml.decode('utf-8'),
```
prabcs: 👍

tests/test_utils.py
```diff
@@ -0,0 +1,124 @@
...
+        <name-node>name-node</name-node>
+    </global>
+    <end name="end" />
+</workflow-app>""".strip()
+        assert len(et.fromstring(xml)), 'Invalid test XML'
+        tests.utils.assert_valid_workflow(xml)
+    assert str(assertion_info.value).strip() == str(
+        '''An XML validation error occurred.
+
+Error: Invalid app definition, org.xml.sax.SAXParseException; lineNumber: 6; columnNumber: 23; '''
```
prabcs: If SAXParser changes its code or we update its version, it may happen that this exact error may not remain valid.
Is it possible to shrink this error with the most relevant search string which you think might stay across versions ?

tests/utils.py
cfournie: This should be evident enough in our code.

tests/pyoozie/test_xml.py
```diff
@@ -69,14 +69,14 @@ def test_workflow_submission_xml_with_configuration(username, workflow_app_path)
...
         <property>
-            <name>other.key</name>
-            <value>other value</value>
-        </property>
-        <property>
             <name>oozie.wf.application.path</name>
             <value>/user/oozie/workflows/descriptive-name</value>
         </property>
         <property>
+            <name>other.key</name>
```
prabcs: 👍

tests/utils.py
```diff
@@ -12,7 +12,7 @@
 import xmltodict
 
 
-def xml_to_dict_unordered(xml):
+def xml_to_comparable_dict(xml):
```
prabcs: 👍

PR #47, opened 2017-05-11
Leave conf None if not provided

kmtaylor-github:
The `fill_in_details()` mechanism relies on detecting that the `conf` field of `Coordinator` and `Workflow` objects was not provided (typical in bulk queries), and does a specific query to fetch the additional information. This change ensures that the parse function preserves `None`.

This is a bug introduced while getting lint to run clean; at one point parsing functions were skipped on null inputs, but the linter saw that as uninitialized variables.

@cfournie @sabidib

### DIFF
--- a/pyoozie/model.py
+++ b/pyoozie/model.py
@@ -91,11 +91,14 @@ def _parse_time(_, time_string):
 
 
 def _parse_configuration(_, conf_string):
-    if conf_string:
+    if conf_string is None:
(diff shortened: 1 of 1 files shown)


sabidib:
👍  This means `fill_in_details` will properly make a request for conf info when you call `.conf`on coordinator that has been retrieved in a Bulk query.

PR #48, opened 2017-06-06
Fix flaky XML comparison

cfournie:
### Problem
When testing https://github.com/Shopify/pyoozie/pull/46, sometimes XML comparisons would fail due to the ordering of lists of dicts (this happens maybe 10% of the time).

```
...
E             'fork': {'@name': 'fork-00000005',
E             -                            'path': [{'@start': 'action-00000000'},
E             ?                                                               ^
E             +                            'path': [{'@start': 'action-00000001'},
E             ?                                                               ^
E             -                                     {'@start': 'action-00000001'}]},
E             ?                                                               ^
E             +                                     {'@start': 'action-00000000'}]},
E             ?                                                               ^
...
```

### Solution
This PR refactors `xml_to_comparable_dict` to convert `collections.OrderedDict` to a `dict` in `_unorder`, and if there's a list, it tries to predictably  sort it by recursing through a value's contents and getting a hashable object for each type in `_sort_key`.

cc/ @Shopify/data-acceleration

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -37,11 +37,20 @@ def valid_workflow():
 def test_xml_to_comparable_dict():
     document_dict = tests.utils.xml_to_comparable_dict("""
 <root>
+
+  <!-- Various tag forms -->
   <tag />
   <tag key="value" />
   <tag>Text</tag>
   <tag key="value">Text</tag>
   <tag different-key="different-value"> Text </tag>
+
+  <!-- Nested dicts -->
+  <fork name="fork-name">
+    <path start="path-1" />
+    <path start="path-2" />
+  </fork>
+
 </root>
     """.strip())
     assert document_dict == {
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -13,14 +13,26 @@
 
 
 def xml_to_comparable_dict(xml):
-    def unorder(value):
+
+    def _sort_key(value):
+        """Recursively sort lists embedded within dicts."""
+        if hasattr(value, 'items'):
+            return six.text_type(sorted([(k, _sort_key(v)) for k, v in value.items()]))
```
prabcs: Is the reason for not using `iteritems` that this can be run in Python 2 as well as 3 ?

tests/test_utils.py
```diff
@@ -37,22 +37,38 @@ def valid_workflow():
...
                 {'#text': 'Text', '@key': 'value'},
                 {'@key': 'value'},
-                'Text',
-            ]
+            ],
+            'fork': {
+                '@name': 'fork-name',
+                'path': [
+                    {'@start': 'path-1'},
+                    {'@start': 'path-2'}
```
prabcs: 👍

prabcs:
LGTM

tests/utils.py
cfournie: Yes

PR #49, opened 2017-06-08
Add basic workflow generation

cfournie:
This PR adds to the `WorkflowApp` class the ability to serialize to XML [action](https://oozie.apache.org/docs/4.1.0/WorkflowFunctionalSpec.html#a3.2_Workflow_Action_Nodes) and [kill](https://oozie.apache.org/docs/4.1.0/WorkflowFunctionalSpec.html#a3.1.3_Kill_Control_Node) nodes using the [`oozie-workflow-0.5.xsd` schema](https://github.com/apache/oozie/blob/master/client/src/main/resources/oozie-workflow-0.5.xsd).

### Implementation
This PR implements:
  - An abstract `_WorkflowEntity` class to serve as the base type for these and other structures that will be exposed to users to allow them to define a workflow which features:
    - An optional `on_error` parameter which allows one to specify a `_WorkflowEntity` to transition to if an error occurs
    - An optional `name` to help document identifiers in the XML (otherwise a uid is used)
  - A `Kill` `_WorkflowEntity` which if transitioned to kills a workflow and emits a message
  - An `Action ` `_WorkflowEntity` which wraps a `ConcreteAction` (e.g. execute a shell command, subworkflow, or email) which
    - Can specify credentials that are to be used by a concrete action
    - Can specify action retry semantics

This PR also adjusts `WorkflowApp` to validate that:
  - All credentials referred to by actions are supplied in the constructor
  - No duplicate identifiers are present in the actions provided (an Oozie logic error)

This PR enables the ability to define collections of actions that can be executed serially/in parallel (https://github.com/Shopify/pyoozie/pull/50).

🎩 'd by starting a local [Oozie 4.1.0 instance](https://github.com/cfournie/docker-hadoop/tree/oozie/2.6.0) using `docker-compose up -d` and manipulating `tests/pyoozie/test_interactive.py` to specify different values.

#### Notes

### DIFF
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -496,7 +496,7 @@ class CoordinatorApp(XMLSerializable):
         return doc
 
 
-class _WorkflowEntity(typing.Iterable):
+class _AbstractWorkflowEntity(typing.Iterable):
     """An abstract object representing an Oozie workflow action that can be serialized to XML."""
     # pylint: disable=abstract-method
 
@@ -506,7 +506,7 @@ class _WorkflowEntity(typing.Iterable):
             self,
             xml_tag,       # typing.Text
             name=None,     # type: typing.Optional[typing.Text]
-            on_error=None  # type: typing.Optional[_WorkflowEntity]
+            on_error=None  # type: typing.Optional[_AbstractWorkflowEntity]
     ):
         # type: (...) -> None
         self.__xml_tag = xml_tag
@@ -538,7 +538,7 @@ class _WorkflowEntity(typing.Iterable):
         raise NotImplementedError()
 
     def __iter__(self):
-        # type: () -> typing.Generator[_WorkflowEntity, None, None]
+        # type: () -> typing.Generator[_AbstractWorkflowEntity, None, None]
         yield self
         if self.__on_error:
             for action in self.__on_error:
@@ -554,7 +554,7 @@ class _WorkflowEntity(typing.Iterable):
         return str('<{_class}({identifier})>'.format(_class=type(self).__name__, identifier=self.identifier()))
 
 
-class Kill(_WorkflowEntity):
+class Kill(_AbstractWorkflowEntity):
     """Workflow graph terminal node(s) to end upon to indicate failure."""
 
     def __init__(self, message, name=None):
@@ -573,7 +573,7 @@ class Kill(_WorkflowEntity):
 ConcreteAction = typing.Union[Shell, SubWorkflow, Email]
 
 
-class Action(_WorkflowEntity):
+class Action(_AbstractWorkflowEntity):
     """Workflow action nodes carrying concrete actions that perform an action."""
 
     def __init__(
@@ -583,7 +583,7 @@ class Action(_WorkflowEntity):
             credential=None,      # type: typing.Optional[typing.Text]
             retry_max=None,       # type: typing.Optional[int]
             retry_interval=None,  # type: typing.Optional[int]
-            on_error=None         # type: typing.Optional[_WorkflowEntity]
+            on_error=None         # type: typing.Optional[_AbstractWorkflowEntity]
     ):
         # type: (...) -> None
         super(Action, self).__init__(xml_tag='action', name=name, on_error=on_error)
@@ -630,7 +630,7 @@ class WorkflowApp(XMLSerializable):
             job_tracker=None,             # type: typing.Optional[typing.Text]
             name_node=None,               # type: typing.Optional[typing.Text]
             job_xml_files=None,           # type: typing.Optional[JobXmlFilesType]
-            entities=None,                # type: typing.Optional[_WorkflowEntity]
+            entities=None,                # type: typing.Optional[_AbstractWorkflowEntity]
     ):
         # type: (...) -> None
         XMLSerializable.__init__(self, 'workflow-app')
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -551,7 +551,7 @@ class _WorkflowEntity(typing.Iterable):
         return self.__bool__()
 
     def __repr__(self):  # type: () -> str
-        return str('{_class}({identifier})'.format(_class=type(self).__name__, identifier=self.identifier()))
+        return str('<{_class}({identifier})>'.format(_class=type(self).__name__, identifier=self.identifier()))
 
 
 class Kill(_WorkflowEntity):
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -640,8 +640,8 @@ class WorkflowApp(XMLSerializable):
             job_xml_files=job_xml_files,
             configuration=configuration
         )
-        self.__credentials = credentials or []
-        self.__actions = copy.deepcopy(actions)
+        self.__credentials = copy.deepcopy(credentials) or []
+        self.__actions = copy.deepcopy(actions) or []
         self.__validate()
 
     def __validate(self):  # type () -> None
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -607,9 +607,9 @@ class Action(_WorkflowEntity):
         }
         if self.__credential:
             attributes['cred'] = self.__credential
-        if self.__retry_max:
+        if self.__retry_max is not None:
             attributes['retry-max'] = str(self.__retry_max)
-        if self.__retry_interval:
+        if self.__retry_interval is not None:
             attributes['retry-interval'] = str(self.__retry_interval)
         with tag(self.xml_tag(), **attributes):
             self.__action._xml(doc, tag, text)
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -547,6 +547,12 @@ class _WorkflowEntity(typing.Iterable):
             for action in self.__on_error:
                 yield action
 
+    def __bool__(self):  # type: () -> bool
+        return bool(list(self))
+
+    def __nonzero__(self):  # type: () -> bool
+        return self.__bool__()
+
     def __repr__(self):
         # type: () -> str
         return str('{_class}({identifier})'.format(_class=type(self).__name__, identifier=self.identifier()))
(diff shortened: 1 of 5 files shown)


pyoozie/tags.py
```diff
@@ -505,35 +624,68 @@ def __init__(
...
         # type: (...) -> None
         XMLSerializable.__init__(self, 'workflow-app')
-        self.name = validate_xml_name(name)
 
-        self.parameters = Parameters(parameters)
-        self.global_configuration = GlobalConfiguration(
+        # XML-document-related values
+        self.__name = validate_xml_name(name)
+        self.__parameters = Parameters(parameters)
+        self.__global_configuration = GlobalConfiguration(
```
cfournie: Let's make these private while we're at it, once we're done validation, these are our objects and should not be manipulated later.

pyoozie/tags.py
```diff
@@ -505,35 +624,68 @@ def __init__(
...
+
+        def _parse_action(action):
+            action_identifiers.append(action.identifier())
+            if hasattr(action, 'credential'):
+                credential = action.credential()
+                if credential:
+                    credentials_needed.add(credential)
+
+        if self.__actions:
+            for action in set(self.__actions):
```
cfournie: In future PRs these iterators may return multiple instances of the same object since they are performing graph traversal and an action's OK and ERROR conditions could both point to the same object.

pyoozie/tags.py
```diff
@@ -494,6 +496,133 @@ def _xml(self, doc, tag, text):
...
+            self,
+            xml_tag,       # typing.Text
+            name=None,     # type: typing.Optional[typing.Text]
+            on_error=None  # type: typing.Optional[_WorkflowEntity]
+    ):
+        # type: (...) -> None
+        self.xml_tag = xml_tag
+        self.__name = name
+        self.__on_error = copy.deepcopy(on_error)
+        self.__uid = uuid.uuid4().hex[:8]
```
prabcs: As `__uid` is only used during creation of the identifier, we can assign `__name` as - 
```self.__name = name or uuid.uuid4().hex[:8]```. This will help make the `create_identifier` method succinct.

File: pyoozie/tags.py [hunk 4]
```diff
@@ -494,6 +496,133 @@ def _xml(self, doc, tag, text):
         return doc
 
 
+class _WorkflowEntity(typing.Iterable):
```
honkfestival: Why not `_AbstractWorkflow`?

pyoozie/tags.py
```diff
@@ -505,35 +634,69 @@ def __init__(
...
+        self.__parameters = Parameters(parameters)
+        self.__global_configuration = GlobalConfiguration(
             job_tracker=job_tracker,
             name_node=name_node,
             job_xml_files=job_xml_files,
             configuration=configuration
         )
-        self.credentials = credentials
+        self.__credentials = credentials or []
+        self.__actions = copy.deepcopy(actions)
```
honkfestival: It seems you should be hangling both of these consistently, i.e.

```py
self.__credentials = copy.deepcopy(credentials) or []
self.__actions = copy.deepcopy(actions) or []
```

Am I missing something?

pyoozie/tags.py
```diff
@@ -494,6 +496,133 @@ def _xml(self, doc, tag, text):
...
+    def _xml(self, doc, tag, text, on_next, on_error):
+        # type: (yattag.doc.Doc, yattag.doc.Doc.tag, yattag.doc.Doc.text, typing.Text, typing.Text) -> yattag.doc.Doc
+        _on_error = self._xml_and_get_on_error(doc, tag, text, on_next, on_error)
+
+        attributes = {
+            'name': self.identifier(),
+        }
+        if self.__credential:
+            attributes['cred'] = self.__credential
+        if self.__retry_max:
```
prabcs: What happens if there is a `__retry_max` but no `__retry_interval` ?

pyoozie/tags.py
```diff
@@ -505,35 +634,69 @@ def __init__(
...
+        self.__actions = copy.deepcopy(actions)
+        self.__validate()
+
+    def __validate(self):  # type () -> None
+        action_identifiers = []
+        credentials_needed = set()
+
+        def _parse_action(action):
+            action_identifiers.append(action.identifier())
+            if hasattr(action, 'credential'):
```
prabcs: Why is this explicit check required ?
If this is required, why not in the previous line to check for `identifier` in `action` object ?

pyoozie/tags.py
```diff
@@ -494,6 +496,133 @@ def _xml(self, doc, tag, text):
...
+            on_error=None  # type: typing.Optional[_WorkflowEntity]
+    ):
+        # type: (...) -> None
+        self.xml_tag = xml_tag
+        self.__name = name
+        self.__on_error = copy.deepcopy(on_error)
+        self.__uid = uuid.uuid4().hex[:8]
+        self.__identifier = self.create_identifier(xml_tag)
+
+    def create_identifier(self, xml_tag):
```
honkfestival: This function should be private, unless you have a use case in mind for users creating identifiers to be used somewhere.

pyoozie/tags.py
```diff
@@ -494,6 +496,133 @@ def _xml(self, doc, tag, text):
...
+    def _xml(self, doc, tag, text, on_next, on_error):
+        # type: (yattag.doc.Doc, yattag.doc.Doc.tag, yattag.doc.Doc.text, typing.Text, typing.Text) -> yattag.doc.Doc
+        raise NotImplementedError()
+
+    def __iter__(self):
+        # type: () -> typing.Generator[_WorkflowEntity, None, None]
+        yield self
+        if self.__on_error:
+            for action in self.__on_error:
+                yield action
```
honkfestival: You could probably make this shorter with `itertools.chain`, but it's pretty clear the way it is. 👍

pyoozie/tags.py
```diff
@@ -494,6 +496,133 @@ def _xml(self, doc, tag, text):
...
+
+    def __bool__(self):  # type: () -> bool
+        return bool(list(self))
+
+    def __nonzero__(self):  # type: () -> bool
+        return self.__bool__()
+
+    def __repr__(self):  # type: () -> str
+        return str('{_class}({identifier})'.format(_class=type(self).__name__, identifier=self.identifier()))
```
honkfestival: This is nitpicky, but could you please surround the return value with `<` and `>`? The string you're building here looks like something you could paste into a REPL to get back an equivalent object, but I'm guessing it doesn't actually round-trip due to the mismatch between the constructor parameter `xml_tag` and the `identifier` you're using here.

Motivation:

> For many types, this function makes an attempt to return a string that would yield an object with the same value when passed to eval(), otherwise the representation is a string enclosed in angle brackets that contains the name of the type of the object together with additional information often including the name and address of the object.

You could of course create a representation that _does_ round-trip, but that would make your identifier generation code a little more complicated.

pyoozie/tags.py
```diff
@@ -494,6 +496,133 @@ def _xml(self, doc, tag, text):
...
+
+    def _xml(self, doc, tag, text, on_next, on_error):
+        # type: (yattag.doc.Doc, yattag.doc.Doc.tag, yattag.doc.Doc.text, typing.Text, typing.Text) -> yattag.doc.Doc
+        with tag(self.xml_tag, name=self.identifier()):
+            with tag('message'):
+                doc.text(self.message)
+        return doc
+
+
+ConcreteAction = typing.Union[Shell, SubWorkflow, Email]
```
honkfestival: 👍

pyoozie/tags.py
cfournie: `_AbstractWorkflowEntity` would be more precise (since these are execution control nodes contained within a workflow), but it's also really verbose. What is the convention for this? I can find none in the [Google styleguide](https://google.github.io/styleguide/pyguide.html) or [`shopify_python`](https://github.com/Shopify/shopify_python).

pyoozie/tags.py
cfournie: All entity identifiers need to be checked to find duplicates, but not all entities have credentials. I should probably rename action to entity.

pyoozie/tags.py
cfournie: Not sure, I'll have to test that.

pyoozie/tags.py
cfournie: Oh neat, I missed the angle brackets part. I'll add that 👍

pyoozie/tags.py
cfournie: Tested:
- If either or both are unspecified, Oozie will use [default values](url) (defined as `oozie.action.retries.max` or `oozie.action.retry.interval` in 4.2.0)
- If negative values are specified, they are accepted (unsure of how they're interpreted though but a workflow can complete successfully).

tests/pyoozie/test_interactive.py
```diff
@@ -0,0 +1,87 @@
...
+import datetime
+import os
+
+import pyoozie
+import pytest
+import pywebhdfs.webhdfs
+import six
+
+
+@pytest.mark.skipif(not bool(os.environ.get(str('INTERACTIVE'))), reason='Requires INTERACTIVE=1 env var')
```
cfournie: This test isn't run unless this env var is present because Travis CI doesn't start up an Oozie server yet (and this test is meant to be interactive, i.e. it will pause for a response before cleaning up its coordinator created on Oozie).

pyoozie/tags.py
cfournie: Function is now gone

pyoozie/tags.py
cfournie: For this iteration `actions` (i.e. now entities) is not a collection, so they'll be treated a little differently to start. https://github.com/Shopify/pyoozie/pull/50 will bring in entity collections and then these can be treated more similarly.

pyoozie/tags.py
cfournie: Just went with `_AbstractWorkflowEntity`. It's not too long.

cfournie:
@honkfestival @prabcs @sabidib @solackerman any final comments?

PR #50, opened 2017-06-08
Add collections of actions to be executed serially/in parallel

cfournie:
This PR adds to the `WorkflowApp` class the ability to define collections of actions to execute either serially or in parallel (using [`fork` and `join` tags](https://oozie.apache.org/docs/4.1.0/WorkflowFunctionalSpec.html#a3.1.5_Fork_and_Join_Control_Nodes) defined in the [`oozie-workflow-0.5.xsd` schema](https://github.com/apache/oozie/blob/master/client/src/main/resources/oozie-workflow-0.5.xsd)) and allows for error-handling to be defined for an entire collection (and its nested contents).

## Implementation
This API allows a user to define a workflow using `Serial` objects to chain actions together and to fork/join a workflow's execution using `Parallel` objects. An example of how these objects can be nested within each other and used to specify `on_error` conditions is shown below.

```
pyoozie.WorkflowApp(
    name='descriptive-name',
    job_tracker='job-tracker',
    name_node='name-node',
    entities=pyoozie.Parallel(
        pyoozie.Action(tags.Shell(exec_command='echo', arguments=['build'])),
        pyoozie.Action(tags.Shell(exec_command='echo', arguments=['resolve'])),
        on_error=pyoozie.Serial(
            pyoozie.Action(tags.Shell(exec_command='echo', arguments=['error'])),
            pyoozie.Kill('A bad thing happened')
        )
    )
)
```

When serialized to XML and submitted to Oozie, this is interpreted as a directed acyclic execution graph.

graph

This nested data structure along with a choice to make these objects as immutable as is practical was chosen so that the API would not provide ways to create illegal graphs, e.g. those with:
  - Cycles;
  - Unmatched fork and join nodes (taken care of by a single `Parallel` object);
  - Paths that do not terminate (i.e. end at either an end or kill node).

This API does not prevent accidentally creating cycles or node identifier collisions, but thankfully duplicate id checking is added in https://github.com/Shopify/pyoozie/pull/49 (the PR upon which this depends).

### Error handling
This PR also demonstrates `on_error` ability to specify error handling in a parent and have children inherit it, e.g.:

```
workflow_app = tags.WorkflowApp(
    name='descriptive-name',
    job_tracker='job-tracker',
    name_node='name-node',
    entities=tags.Serial(
        tags.Parallel(
            tags.Action(name='build_a', action=tags.Shell(exec_command='echo', arguments=['build_a'])),
            tags.Action(name='build_b', action=tags.Shell(exec_command='echo', arguments=['build_b'])),
            name='builders'
        ),
        tags.Parallel(
            tags.Action(name='resolve_a', action=tags.Shell(exec_command='echo', arguments=['resolve_a'])),
            tags.Action(name='resolve_b', action=tags.Shell(exec_command='echo', arguments=['resolve_b'])),
            name='resolvers'
        ),
        on_error=tags.Serial(
            tags.Action(name='error', action=tags.Shell(exec_command='echo', arguments=['error'])),
            tags.Kill(name='error', message='A bad thing happened')
        )
    )
)
```
The error handling is defined in a containing `Serial` class, and the child actions within the two `Parallel`s all have their error transitions routed to the `Serial`'s error-handler.

graph2

#### Notes


pyoozie/tags.py
```diff
@@ -613,6 +615,87 @@ def _xml(self, doc, tag, text, on_next, on_error):
         return doc
 
 
+class Serial(_WorkflowEntity):
+    """Sequence of actions to execute (implemented by chaining actions and 'OK' transitions)"""
+
+    def __init__(self, *actions, **kwargs):
+        # type: (*_WorkflowEntity, **_WorkflowEntity) -> None
+        super(Serial, self).__init__(xml_tag=None, on_error=kwargs.get(str('on_error')))
+        self.__actions = tuple(copy.deepcopy(actions))  # type: typing.Tuple[_WorkflowEntity, ...]
```
kmtaylor-github: This is something we're going to have to document carefully. Adding an action to a `Serial` / `Parallel` makes it very hard to adjust the action afterwards, so certain "obvious" patterns will be wrong.
Example:
 ❌   create 3 jobs, run in parallel, create 3 notifications & attach to jobs
 ✅   create 3 jobs, create 3 notifications & attach to jobs, run in parallel

tests/pyoozie/test_tags.py
```diff
@@ -586,3 +589,170 @@ def test_workflow_with_reused_identifier():
...
+    with pytest.raises(AssertionError) as assertion_info:
+        tags.WorkflowApp(
+            name='descriptive-name',
+            job_tracker='job-tracker',
+            name_node='name-node',
+            actions=tags.Serial(
+                tags.Action(name='build', action=tags.Shell(exec_command='echo', arguments=['build'])),
+                tags.Action(name='resolve', action=tags.Shell(exec_command='echo', arguments=['resolve'])),
+                on_error=tags.Serial(
+                    tags.Action(name='build', action=tags.Shell(exec_command='echo', arguments=['error'])),
```
kmtaylor-github: An interesting side question: should `name`s be unique?

tests/pyoozie/test_tags.py
```diff
@@ -586,3 +589,170 @@ def test_workflow_with_reused_identifier():
...
+                    tags.Kill('A bad thing happened')
+                )
+            )
+        )
+    assert str(assertion_info.value) == 'Name(s) reused: action-build'
```
kmtaylor-github: Derp... the answer is clearly YES. 👍 👍

pyoozie/tags.py
```diff
@@ -639,7 +722,7 @@ def __init__(
             configuration=configuration
         )
         self.__credentials = credentials or []
-        self.__actions = actions
+        self.__actions = actions or Serial()
```
cfournie: An empty `Serial` will simply transition to `end`.

tests/pyoozie/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+import uuid
+
+import mock
+import pytest
+
+from six import moves
+
+
+@pytest.fixture
+def fake_uuid4(monkeypatch):
```
cfournie: Let's test our uid identifiers generated in `_WorkflowEntity`, but reduce their randomness a bit.

pyoozie/tags.py
cfournie: Anything that's permitted to be edited can be, but most properties are locked down and most constructors copy. To split up object creation you need to start from the leaves inwards with this API.

pyoozie/tags.py
```diff
@@ -649,7 +722,7 @@ def __init__(
             configuration=configuration
         )
         self.__credentials = credentials or []
-        self.__actions = copy.deepcopy(actions)
+        self.__actions = copy.deepcopy(actions) if actions else Serial()
```
cfournie: An empty Serial will produce no XML (leading to a transition to the end tag).

prabcs:
Is the join [as shown here](https://user-images.githubusercontent.com/49264/26977534-3c94c600-4cf6-11e7-8b3d-be6bc2f395cd.png) a blocking operation on `ok` from all the parallel actions ? I'm trying to comprehend what happens when action 1 (`build` in this case) errors out but action 2 (`resolve`) passes.

kmtaylor-github:
> Is the join as shown here a blocking operation

A `join` is a blocking operation, but a `kill` terminates the entire flow. In the pictured example if either node transitions to error, the other will be killed.

This is often undesirable, so the transition to `kill` is delayed until the `join` completed. See https://github.com/Shopify/pyoozie/pull/51 for details.

pyoozie/tags.py
solackerman: this `self.__actions` is a `tuple`, but the parallel one is a `frozen_set`. Is that to make it clear that order does not matter for the parallel case, but it does for the serial?

pyoozie/tags.py
```diff
@@ -623,6 +624,78 @@ def _xml(self, doc, tag, text, on_next, on_error):
...
+
+
+class Parallel(_WorkflowEntity):
+    """Set of actions to execute in parallel (implemented as fork/join tag pair)"""
+
+    def __init__(self, *actions, **kwargs):
+        # type: (*_WorkflowEntity, **typing.Union[_WorkflowEntity, typing.Text]) -> None
+        name = kwargs.get(str('name'))
+        name = name if isinstance(name, six.string_types) else None
+        on_error = kwargs.get(str('on_error'))
```
solackerman: why is this `str('on_error')` instead of `'on_error'`?

pyoozie/tags.py
```diff
@@ -623,6 +624,78 @@ def _xml(self, doc, tag, text, on_next, on_error):
...
+        on_error = on_error if isinstance(on_error, _WorkflowEntity) else None
+        super(Parallel, self).__init__(
+            xml_tag=None,
+            name=name,
+            on_error=on_error
+        )
+        assert actions, 'At least 1 action required'
```
solackerman: Can we move this assertion up to the top? I don't see the point in waiting.

tests/pyoozie/test_tags.py
```diff
@@ -619,3 +622,187 @@ def test_workflow_with_reused_identifier():
...
+                name='resolvers'
+            ),
+            on_error=tags.Serial(
+                tags.Action(name='error', action=tags.Shell(exec_command='echo', arguments=['error'])),
+                tags.Kill(name='error', message='A bad thing happened')
+            )
+        )
+    )
+
+    assert_workflow(request, workflow_app)
```
solackerman: Why are we changing approaches for this test, and instead of testing for an `expected_xml`, testing that just the `on_error`'s work as expected?

pyoozie/tags.py
cfournie: > Is that to make it clear that order does not matter for the parallel case, but it does for the serial?

Correct 😄

tests/pyoozie/test_tags.py
cfournie: The XML for this is huge, so this is checking that the XML validates against the schema, nothing more. Below we then test some properties about the paths that should be defined within the XML.

pyoozie/tags.py
```diff
@@ -504,26 +505,25 @@ class _AbstractWorkflowEntity(typing.Iterable):
...
 
     def xml_tag(self):
         # type: () -> typing.Text
         return self.__xml_tag
 
+    def identifier(self, xml_tag=None):
+        # type: (typing.Optional[typing.Text]) -> typing.Text
+        if not xml_tag:
+            xml_tag = self.__xml_tag
+        return validate_xml_id('{tag}-{name}'.format(tag=xml_tag, name=self.__name))
```
cfournie: Refactored because for the "parallel" collection we need to actually create two identifiers, but we want them to share the same "name", e.g. `fork-0001` and `join-0001` are two different identifiers for two different tags, but we want to be able to relate that these tags are linked together (by sharing the name `0001`).

pyoozie/tags.py
```diff
@@ -619,6 +619,73 @@ def _xml(self, doc, tag, text, on_next, on_error):
...
+class Serial(_AbstractWorkflowEntity):
+    """Sequence of entities to execute (implemented by chaining entities and 'OK' transitions)"""
+
+    def __init__(self, *entities, **kwargs):
+        # type: (*_AbstractWorkflowEntity, **_AbstractWorkflowEntity) -> None
+        super(Serial, self).__init__(on_error=kwargs.get(str('on_error')))
+        self.__entities = tuple(copy.deepcopy(entities))  # type: typing.Tuple[_AbstractWorkflowEntity, ...]
+
+    def identifier(self, xml_tag=None):  # type: (typing.Optional[typing.Text]) -> typing.Text
+        return self.__entities[0].identifier(xml_tag) if self.__entities else None
```
cfournie: This is just a sequence, so let's address it by the entry point for this sequence, the first real tag.

pyoozie/tags.py
```diff
@@ -504,26 +505,25 @@ class _AbstractWorkflowEntity(typing.Iterable):
...
-
-    def identifier(self):
-        # type: () -> typing.Text
-        return self.__identifier
 
     def xml_tag(self):
         # type: () -> typing.Text
         return self.__xml_tag
 
+    def identifier(self, xml_tag=None):
```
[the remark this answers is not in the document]
cfournie: It is doing two different things... Refactored into two methods. `identifier()` is not used to get the identifier for an entity so that you can transition to it and and `create_identifier` can be used to create a similar identifier (used in `Parallel`).

pyoozie/tags.py
```diff
@@ -619,6 +619,73 @@ def _xml(self, doc, tag, text, on_next, on_error):
...
+            if entity is not self:  # Don't return self because this collection doesn't result in a node
+                yield entity
+
+
+class Parallel(_AbstractWorkflowEntity):
+    """Set of entities to execute in parallel (implemented as fork/join tag pair)"""
+
+    def __init__(self, *entities, **kwargs):
+        # type: (*_AbstractWorkflowEntity, **typing.Union[_AbstractWorkflowEntity, typing.Text]) -> None
+        name = kwargs.get(str('name'))
```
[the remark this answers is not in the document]
cfournie: Type checking fails.
```
Checking type annotations...
pyoozie/tags.py:661: error: No overload variant of "get" of "Mapping" matches argument types [builtins.unicode]
pyoozie/tags.py:663: error: No overload variant of "get" of "Mapping" matches argument types [builtins.unicode]
```

PR #51, opened 2017-06-08
Add decision action

cfournie:
This PR extends the available collections for Oozie workflow actions to include [`decision` nodes](https://oozie.apache.org/docs/4.1.0/WorkflowFunctionalSpec.html#a3.1.4_Decision_Control_Node) (using the [`oozie-workflow-0.5.xsd` schema](https://github.com/apache/oozie/blob/master/client/src/main/resources/oozie-workflow-0.5.xsd)) with error handling definable using an `on_error` parameter.

### Discussion
This PR implements decision nodes and error handling such that we can accomplish one crucial execution pattern: checkpointing.

When a collection of actions are being executed in parallel and one of them transitions to a kill node, that kills the entire workflow (parallel actions and all). The other parallel actions may complete successfully, but if a single one transitions to kill it interrupts them.

| Without checkpoint | With checkpoint |
| --- | --- |
| graph4 | graph3 |

Using the checkpointing pattern, instead of directly transition to kill within a fork, we short-circuit and transition to the join node (allowing other parallel actions to continue) and after that join we can add a decision node to check whether any previous task has encountered an error and then transition to a kill node.

#### Notes

### DIFF
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -636,8 +636,8 @@ class Decision(_AbstractWorkflowEntity):
         super(Decision, self).__init__(xml_tag='decision', name=name, on_error=on_error)
         assert default, 'A default must be supplied'
         assert choices, 'At least one choice required'
-        self.__default = default
-        self.__choices = choices
+        self.__default = copy.deepcopy(default)
+        self.__choices = copy.deepcopy(choices)
 
     def _xml(self, doc, tag, text, on_next, on_error):
         # type: (yattag.doc.Doc, yattag.doc.Doc.tag, yattag.doc.Doc.text, typing.Text, typing.Text) -> yattag.doc.Doc
--- a/pyoozie/tags.py
+++ b/pyoozie/tags.py
@@ -627,10 +627,10 @@ class Decision(_AbstractWorkflowEntity):
 
     def __init__(
             self,
-            default,        # type: _WorkflowEntity
-            choices,        # type: typing.Dict[typing.Text, _WorkflowEntity]
+            default,        # type: _AbstractWorkflowEntity
+            choices,        # type: typing.Dict[typing.Text, _AbstractWorkflowEntity]
             name=None,      # type: typing.Optional[typing.Text]
-            on_error=None,  # type: typing.Optional[_WorkflowEntity]
+            on_error=None,  # type: typing.Optional[_AbstractWorkflowEntity]
     ):
         # type: (...) -> None
         super(Decision, self).__init__(xml_tag='decision', name=name, on_error=on_error)
@@ -659,7 +659,7 @@ class Decision(_AbstractWorkflowEntity):
         return doc
 
     def __iter__(self):
-        # type: () -> typing.Generator[_WorkflowEntity, None, None]
+        # type: () -> typing.Generator[_AbstractWorkflowEntity, None, None]
         for action in self.__choices.values():
             yield action
         yield self.__default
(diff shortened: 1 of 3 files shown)


pyoozie/tags.py
```diff
@@ -615,6 +615,48 @@ def _xml(self, doc, tag, text, on_next, on_error):
...
+        return doc
+
+    def __iter__(self):
+        for action in self.__choices.values():
+            yield action
+        yield self.__default
+        for action in super(Decision, self).__iter__():
+            yield action
+
+
```
kmtaylor-github: Not providing`__bool__` etc.?

cfournie: Now provided in the parent class, `_WorkflowEntity` 😉

pyoozie/tags.py
```diff
@@ -624,6 +624,51 @@ def _xml(self, doc, tag, text, on_next, on_error):
...
+            with tag('switch'):
+                for case, dest in self.__choices.items():
+                    with tag('case', to=dest.identifier()):
+                        doc.text(case)
+                doc.stag('default', to=self.__default.identifier())
+
+        # Write contained actions
+        self.__default._xml(doc, tag, text, on_next, _on_error)
+        for choice in self.__choices.values():
+            choice._xml(doc, tag, text, on_next, _on_error)
```
kmtaylor-github: I feel like you've already made the comment somewhere (another PR?), but: we should be careful around these. It's very possible that the same node may appear in multiple cases and/or also as the default (and/or as an error transition).

cfournie: The identifier uniqueness check will iterate over each case and disallows duplicate identifiers within them, so this API will prevent duplicate nodes within choices. This is an artificial limitation, but one that makes error-handling code much simpler to implement (since it does add node-sharing to paths).

kmtaylor-github: That's fine, but my point was that it would be valid to, for example, have a transition to the error node on both error and a particular switch case.

cfournie: True. Valid Oozie XML, but not enabled here for simplicity.

pyoozie/tags.py
```diff
@@ -624,6 +624,51 @@ def _xml(self, doc, tag, text, on_next, on_error):
...
+
+        return doc
+
+    def __iter__(self):
+        # type: () -> typing.Generator[_WorkflowEntity, None, None]
+        for action in self.__choices.values():
+            yield action
+        yield self.__default
+        for action in super(Decision, self).__iter__():
+            yield action
```
solackerman: is this the standard iterator order in this repo? First leaves, then default, then superclass?

cfournie: It's arbitrary, there shouldn't be anything that depends upon a specific order.

tests/pyoozie/test_tags.py
```diff
@@ -833,3 +833,94 @@ def test_workflow_app_inherit_parent_error(request):
...
     app.assert_node("/action[@name='action-error']/error", to='end')
     app.assert_node("/action[@name='action-error']/ok", to='kill-error')
+
+
+def test_workflow_app_empty_decision_entities():
+    with pytest.raises(AssertionError) as assertion_info:
+        tags.Decision(
+            default=None,
+            choices={
+                '${wf:lastErrorNode() eq null}': tags.Action(
```
sabidib: Are users of this library going to need to know how to express conditionals like this? 

It might make sense to abstract this away.

cfournie: Yes, that's a [JSP-EL](https://docs.oracle.com/javaee/1.4/tutorial/doc/JSPIntro7.html) with Oozie entities. There's enough available variation that I'd like to not make an abstraction for this. Instead, on top of this API, I'd like to provide "recipes" for common problems and provide the required JSP-EL in the future.

PR #53, opened 2017-07-14
Change how we specify environment markers

cfournie:
The latest version of setuptools (36.2.0) introduced https://github.com/pypa/setuptools/pull/1085, which disallows the manner in which we were specifying environment markers, which broke a variety of env marker usages https://github.com/pypa/setuptools/issues/1087.

This PR properly specifies env markers.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -49,7 +49,8 @@ load-plugins=shopify_python
 # fixme; we're in-development, so we need some of these
 disable=missing-docstring,too-few-public-methods,too-many-instance-attributes,too-many-arguments,
  protected-access,invalid-name,redefined-outer-name,no-self-use,too-many-public-methods,too-many-lines,
- duplicate-code,fixme
+ duplicate-code,fixme,no-else-return,not-callable,len-as-condition,cond-expr,useless-super-delegation,
+ super-init-not-called
 
 # Increase max except nodes because we often have a pattern of
 # record stats, log, and then re-raise when handling HTTP exceptions
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ setuplib.setup(
             'pytest-randomly',
             'pytest>=2.7',
             'requests-mock',
-            'shopify_python==0.2.2',
+            'shopify_python==0.4.1',
             'xmltodict',
             'pywebhdfs>=0.4.1',
         ],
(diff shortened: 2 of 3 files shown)


benoit-pierre:
You don't need to name it differently, this would work:
```python
    extras_require={
        'test': ['flake8'],
        'test:python_version >= "3.4"': ['mypy'],
    },
```
installing both `flake8` and `mypy` when using Python 3.

setup.py
```diff
@@ -43,17 +43,19 @@ def get_version():
             'autopep8',
             'flake8',
             'mock',
-            'mypy; python_version >= "3.3"',
             'pycodestyle == 2.2.0',
-            'pylint>=1.6.5,<1.7',
+            'pylint>=1.7.1,<1.8',
```
cfournie: Needed to be in-sync with `shopify_python==0.4.1`

setup.py
```diff
@@ -43,17 +43,19 @@ def get_version():
...
             'pytest-cov',
             'pytest-randomly',
             'pytest>=2.7',
             'requests-mock',
-            'shopify_python==0.2.2',
+            'shopify_python==0.4.1',
```
cfournie: Updated to also fix issue installing with `setuptools` 36.2.0.

pylintrc
```diff
@@ -49,7 +49,8 @@ load-plugins=shopify_python
 # fixme; we're in-development, so we need some of these
 disable=missing-docstring,too-few-public-methods,too-many-instance-attributes,too-many-arguments,
  protected-access,invalid-name,redefined-outer-name,no-self-use,too-many-public-methods,too-many-lines,
- duplicate-code,fixme
+ duplicate-code,fixme,no-else-return,not-callable,len-as-condition,cond-expr,useless-super-delegation,
+ super-init-not-called
```
cfournie: Updating `pylint` and `shopify_python` brings with it new lint rules which we violate. Temporarily ignoring them.

setup.py
```diff
@@ -43,17 +43,19 @@ def get_version():
             'autopep8',
             'flake8',
             'mock',
-            'mypy; python_version >= "3.3"',
```
cfournie: This is the bad env marker in question.

pylintrc
sabidib: We should make sure to open an issue about these and ensure we actually need them ignored.

PR #57, opened 2017-09-18
Suspended coordinators api

lidiyam:
Added a public suspended coordinators api, needed for: https://github.com/Shopify/dogpound/pull/50

### DIFF
--- a/pyoozie/__init__.py
+++ b/pyoozie/__init__.py
@@ -38,7 +38,7 @@ from pyoozie.tags import validate_xml_name
 from pyoozie.tags import WorkflowApp
 
 
-__version__ = '0.0.1'
+__version__ = '0.0.2'
 
 __all__ = (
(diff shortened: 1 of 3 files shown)


cfournie:
> Any other "obvious" variants we're missing?

Those that are in prep? Might show us how many new coordinators were created and are awaiting their first start. That's of dubious usefulness at the moment, no need to add it in this PR.

pyoozie/__init__.py
```diff
@@ -38,7 +38,7 @@
 from pyoozie.tags import WorkflowApp
 
 
-__version__ = '0.0.1'
+__version__ = '0.0.2'
```
cfournie: Normally we create another PR to bump the version (b/c there's usually a backlog of changes to release), but since there's no backlog, we can bump in this PR 👍

PR #59, opened 2017-10-27
[WIP] Add re-run ability

sabidib:


pyoozie/client.py
```diff
@@ -515,6 +515,13 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
             return True
         return False
 
+    def job_coordinator_rerun(self, coordinator_id=None, name=None, user=None):
+        coord = self._fetch_coordinator_or_action(coordinator_id, name, user)
```
kmtaylor-github: Rerun only applies to an action, so you should probably assert if it isn't.

PR #60, opened 2017-11-01
Add job_coordinator_rerun

angelini:
Added `job_coordinator_rerun` which allows us to run the `coord-rerun` action on a coordinator action. Necessary for replicating the logic found here: https://github.com/Shopify/starscream/blob/master/starscream/scheduling/oozie/oozie_utils.py#L114-L125

I've also moved around and renamed `jobs_update_coordinator` since it didn't seem to follow the same pattern as the other functions.

AFAICT, functions starting with `jobs_*` hit the `/jobs` endpoint whereas functions starting with `job_*` hit the `/job` endpoint. Also the action name is typically the suffix of the function name.

### DIFF
--- a/pyoozie/client.py
+++ b/pyoozie/client.py
@@ -483,11 +483,15 @@ class OozieClient(object):
     # Job API - manage coordinator
     # ===========================================================================
 
-    def _coordinator_perform_simple_action(self, coord, action):
+    def _coordinator_perform_simple_action(self, coord, action, **kwargs):
+        suffix = '&' + '&'.join('='.join(t) for t in kwargs.items()) if kwargs else ''
         if coord.is_action():
-            self._put('job/{}?action={}&type=action&scope={}'.format(coord.coordJobId, action, coord.actionNumber))
+            self._put('job/{}?action={}&type=action&scope={}{}'.format(coord.coordJobId,
+                                                                       action,
+                                                                       coord.actionNumber,
+                                                                       suffix))
         else:
-            self._put('job/{}?action={}'.format(coord.coordJobId, action))
+            self._put('job/{}?action={}{}'.format(coord.coordJobId, action, suffix))
 
     def _fetch_coordinator_or_action(self, coordinator_id=None, name=None, user=None):
         coord_id = self._decode_coord_id(coordinator_id, name, user)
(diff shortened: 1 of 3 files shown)


pyoozie/client.py
```diff
@@ -515,6 +519,36 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
             return True
         return False
 
+    def job_coordinator_rerun(self, coordinator_id):
+        coord = self._fetch_coordinator_or_action(coordinator_id)
+        if not coord.is_action():
+            raise ValueError('Rerun only supports coodinator action IDs')
```
kmtaylor-github: `s/coodinator/coordinator/`

pyoozie/client.py
```diff
@@ -483,11 +483,15 @@ def job_action_info(self, job_id):
     # Job API - manage coordinator
     # ===========================================================================
 
-    def _coordinator_perform_simple_action(self, coord, action):
+    def _coordinator_perform_simple_action(self, coord, action, **kwargs):
+        suffix = '&' + '&'.join('='.join(t) for t in kwargs.items()) if kwargs else ''
```
kmtaylor-github: 👍

pyoozie/client.py
```diff
@@ -515,6 +519,36 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
             return True
         return False
 
+    def job_coordinator_rerun(self, coordinator_id):
+        coord = self._fetch_coordinator_or_action(coordinator_id)
+        if not coord.is_action():
+            raise ValueError('Rerun only supports coodinator action IDs')
+
+        if not coord.status.is_active():
```
kmtaylor-github: We should also check that the coordinator itself IS active (i.e. `coord.coordinator.status.is_active()`); rerun on a killed coord does not work (and leads to the mysterious `DONEWITHERROR`).

See https://oozie.apache.org/docs/4.1.0/DG_CoordinatorRerun.html

pyoozie/client.py
```diff
@@ -515,6 +519,36 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
...
+            self.logger.info('Rerunning coordinator action %s', coordinator_id)
+            self._coordinator_perform_simple_action(coord, 'coord-rerun', refresh='true')
+            return True
+        return False
+
+    def job_coordinator_update(self, coordinator_id, xml_path, configuration=None):
```
kmtaylor-github: `jobs_` -> `job_`  👍

pyoozie/client.py
```diff
@@ -515,6 +519,36 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
             return True
         return False
 
+    def job_coordinator_rerun(self, coordinator_id):
+        coord = self._fetch_coordinator_or_action(coordinator_id)
+        if not coord.is_action():
+            raise ValueError('Rerun only supports coodinator action IDs')
+
+        if coord.status.is_active():
```
kmtaylor-github: Thanks github. Take 2:
> We should also check that the coordinator itself IS active (i.e. `coord.coordinator.status.is_active()`); rerun on a killed coord does not work (and leads to the mysterious `DONEWITHERROR`).
>
> See https://oozie.apache.org/docs/4.1.0/DG_CoordinatorRerun.html

pyoozie/client.py
```diff
@@ -515,6 +519,36 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
...
+            self.logger.info('Rerunning coordinator action %s', coordinator_id)
+            self._coordinator_perform_simple_action(coord, 'coord-rerun', refresh='true')
```
sabidib: Did you give this a test on an oozie instance?

pyoozie/client.py
```diff
@@ -515,6 +519,36 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
...
+            return True
+        return False
```
sabidib: Let's test the case when the coord status is not active.

pyoozie/client.py
```diff
@@ -515,6 +519,36 @@ def job_coordinator_kill(self, coordinator_id=None, name=None, user=None):
             return True
         return False
 
+    def job_coordinator_rerun(self, coordinator_id):
+        coord = self._fetch_coordinator_or_action(coordinator_id)
```
kmtaylor-github: Style comment, but this might read better by calling this `action`.

angelini:
Tested against a live Oozie instance

PR #64, opened 2018-05-15
Delay the connection check until the first request

### LINKED ISSUE #63: OozieClient constructor should not perform actual network communication
When instantiating a `OozieClient`, the `url` is checked by performing a request to Oozie. 

Code: https://github.com/Shopify/pyoozie/blob/master/pyoozie/client.py#L81

Constructors usually don't perform actual work, so it's counter-intuitive and can make testing harder than it should be.

pior:
# Why

The `OozieClient` constructor is performing a connection check. 
This can make testing harder than it should be.

# How

- Add an argument to disable the check in the constructor
- Make the `test_connection` method part of the API

Resolves #63

### DIFF
--- a/tests/pyoozie/test_client.py
+++ b/tests/pyoozie/test_client.py
@@ -35,8 +35,7 @@ def oozie_config():
 @pytest.fixture
 def api(oozie_config):
     with mock.patch('pyoozie.client.OozieClient._test_connection'):
-        api = client.OozieClient(**oozie_config)
-    return api
+        yield client.OozieClient(**oozie_config)
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


pior:
@kmtaylor-github I don't think that expecting users to mock internals of `OozieClient` to test their own application is reasonable.

Beyond this, a constructor usually only initialize the object and should never fail for external reasons.

pior:
I'm not sure why CI is failing on unchanged code...

kmtaylor-github:
> I don't think that expecting users to mock internals of OozieClient to test their own application is reasonable.

And I don't think users should be testing a client library at all. I still don't understand your case.

pior:
We are not testing PyOozie per se, but we would still need to mock `OozieClient._test_connection` in our functional tests, even though we don't actually perform any actions to Oozie.

pior:
I can mock.patch the client or it's internals.

With this PR, my intention was mostly about slightly improving `OozieClient` by:

- optionally having a constructor that do not fail for external condition
- kind of documenting this _connection check_  with the `test_connection=True` argument.

pior:
@cfournie My comment was misleading, there is nothing else that would need to be mocked.

kmtaylor-github:
> There's likely another situation where a user might want to delay the connection test until a later point

That seems plausible, but I think a cleaner way to accomplish that would be to lazily test the connection on the first request (rather than allowing calls without the check). Something like:
```python
    def _request(self, method, endpoint, content_type, content=None):
        if not self._valid_server:
            self._test_connection()
            self._valid_server = True
        ...
```

That only kicks the can a short way down the road however, as any API is likely to boil down to a `_request` call.

tests/pyoozie/test_client.py
```diff
@@ -151,31 +150,30 @@ class TestOozieClientCore(object):
...
-                m.get('http://localhost:11000/oozie/versions', status_code=404)
-                client.OozieClient(**oozie_config)
-        assert 'Unable to contact Oozie server' in str(err)
-
-        with pytest.raises(exceptions.OozieException) as err:
-            with requests_mock.mock() as m:
-                m.get('http://localhost:11000/oozie/versions', text='>>> fail <<<')
-                client.OozieClient(**oozie_config)
-        assert 'Invalid response from Oozie server' in str(err)
+            client.OozieClient(**oozie_config).admin_build_version()
```
kmtaylor-github: Why not just call `_test_connection` directly?

tests/pyoozie/test_client.py
```diff
@@ -155,26 +155,34 @@ def test_construction(self, mock_test_conn, oozie_config):
...
 
+    def test_test_connection_is_called_once(self, oozie_config):
+        with requests_mock.mock() as m:
+            m.get('http://localhost:11000/oozie/v2/admin/build-version', text='{}')
+
+            with mock.patch('pyoozie.client.OozieClient._test_connection') as m_test:
+                oozie_client = client.OozieClient(**oozie_config)
+                oozie_client.admin_build_version()
+                oozie_client.admin_build_version()
+                m_test.assert_called_once_with()
```
kmtaylor-github: Nice 👍

pior:
CI is failing on the typing checks. Tests are passing successfully.

PR #72, opened 2019-04-02
Allow specifying details flag to avoid extra REST calls

sebmartin:
Surface the `details` flag to avoid making additional REST calls (1 per item in the list) when the extra details are not necessary.

### DIFF
--- a/tests/pyoozie/test_client.py
+++ b/tests/pyoozie/test_client.py
@@ -463,10 +463,10 @@ class TestOozieClientJobsQuery(object):
             mock_query.return_value = [sample_workflow_running]
 
             api.jobs_all_active_workflows()
-            mock_query.assert_called_with(model.ArtifactType.Workflow, user=None, status=expected_statuses)
+            mock_query.assert_called_with(model.ArtifactType.Workflow, details=True, user=None, status=expected_statuses)
 
             api.jobs_all_active_workflows(user='john_doe')
-            mock_query.assert_called_with(model.ArtifactType.Workflow, user='john_doe', status=expected_statuses)
+            mock_query.assert_called_with(model.ArtifactType.Workflow, details=True, user='john_doe', status=expected_statuses)
 
     def test_jobs_all_running_workflows(self, api, sample_workflow_running):
         expected_statuses = model.WorkflowStatus.running()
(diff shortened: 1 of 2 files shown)


cfournie:
Lines need to be a bit shorter, otherwise LGTM

sebmartin:
@kmtaylor-github I bumped the batch size to 5000 as discussed.  Is [this](https://github.com/Shopify/pyoozie/pull/72/commits/ff74a46e5dbc28e4aaaf3f1ec7859214994dc2eb) what you had in mind?

sebmartin:
Remaining/existing CI failures are handled in a [separate PR](https://github.com/Shopify/pyoozie/pull/73).

tests/pyoozie/test_client.py
```diff
@@ -399,8 +399,38 @@ def test_jobs_query_coordinator_pagination(self, _, api):
...
+            ]
+        )
+
+        with mock.patch.object(api, '_get') as mock_get:
+            mock_get.side_effect = lambda url: next(mock_results)
+            api._jobs_query(model.ArtifactType.Coordinator, limit=5)
+            mock_get.assert_called_with('jobs?jobtype=coordinator&offset=1&len=5')
+            api._jobs_query(model.ArtifactType.Coordinator, limit=6000)
+            mock_get.assert_any_call('jobs?jobtype=coordinator&offset=1&len=5000')
+            mock_get.assert_any_call('jobs?jobtype=coordinator&offset=5001&len=5000')
```
kmtaylor-github: Is this test substantially different from / combinable with `test_jobs_query_coordinator_pagination`?
https://github.com/Shopify/pyoozie/blob/045fe6f04ec7f5abbae1a8567feb1fb7bb87ddd4/tests/pyoozie/test_client.py#L383-L405

sebmartin: The difference is testing pagination with and without an explicit limit.  It's pretty subtle but the previous version of the limit code would effectively disable pagination by setting the chunk to be equal to the limit.  This meant that if you set a limit of 20k then you'll request one chunk of 20k instead of X chunks of 500 up to the limit of 20k total results.

[The change that I'm introducing](https://github.com/Shopify/pyoozie/pull/72/files#diff-474ab6c8f76e0a89751359f5b7e19f1dR242) will make sure that even a large limit will be paginated.  This test checks for that by setting a limit of `6000` and checking that two REST calls were made with `len=5000` (the new chunk size).

PR #73, opened 2019-04-10
Fix CI errors on master

sebmartin:
`master` has been failing for a while with type errors.  This attempts to fix them.

https://travis-ci.org/Shopify/pyoozie/jobs/380235270#L971-L1034

Also mypy stopped supporting python 3.4 [about two weeks ago](https://github.com/python/mypy/pull/6592).


sebmartin:
> What are your thoughts on changing this slightly with #74?

Yeah I like that better 👍

PR #74, opened 2019-04-10
Make xml_tag required

cfournie:
Alternate take on https://github.com/Shopify/pyoozie/pull/73 which makes `xml_tag` required to avoid needing to offer default values.


sebmartin:
This was cherry-picked into #73

PR #81, opened 2019-09-09
Replace requests convenience methods with session object 

evadb:
This PR adds a requests.Session() object to OozieClient and replaces the requests convenience methods with the session. This allows OozieClient to pool connections and make more effective use of server resources. 

This will also allow users to pass in their own customized sessions if required.

### DIFF
--- a/pyoozie/client.py
+++ b/pyoozie/client.py
@@ -68,7 +68,7 @@ class OozieClient(object):
         def elapsed(self):
             return self._elapsed
 
-    def __init__(self, url=None, user=None, timeout=None, verbose=True, **_):
+    def __init__(self, url=None, user=None, timeout=None, verbose=True, session=None, **_):
         self.logger = logging.getLogger('pyoozie.OozieClient')
         oozie_url = (url or 'http://localhost').rstrip('/')
         if not oozie_url.endswith('/oozie'):
@@ -79,11 +79,12 @@ class OozieClient(object):
         self._verbose = verbose  # Note: change default for verbose!
         self._stats = OozieClient.Stats()
         self._valid_server = False
+        self._session = session or requests.Session()
 
     def _test_connection(self):
         response = None
         try:
-            response = requests.get('{}/versions'.format(self._url), timeout=self._timeout)
+            response = self._session.get('{}/versions'.format(self._url), timeout=self._timeout)
             response.raise_for_status()
             self._stats.update(response)
         except requests.RequestException as err:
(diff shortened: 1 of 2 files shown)


pyoozie/client.py
```diff
@@ -79,11 +79,14 @@ def __init__(self, url=None, user=None, timeout=None, verbose=True, **_):
         self._verbose = verbose  # Note: change default for verbose!
         self._stats = OozieClient.Stats()
         self._valid_server = False
+        if not type(session) == requests.sessions.Session:
```
kmtaylor-github: Use `isinstance` in case there are subclasses of `Session`?

tests/pyoozie/test_client.py
```diff
@@ -188,6 +212,7 @@ def test_request(self, api):
             m.get('http://localhost:11000/oozie/v2/endpoint', text='{"result": "pass"}')
             result = api._request('GET', 'endpoint', None, None)
             assert result['result'] == 'pass'
+
```
kmtaylor-github: 🔥

pyoozie/client.py
```diff
@@ -79,11 +79,14 @@ def __init__(self, url=None, user=None, timeout=None, verbose=True, **_):
         self._verbose = verbose  # Note: change default for verbose!
         self._stats = OozieClient.Stats()
         self._valid_server = False
+        if not type(session) == requests.sessions.Session: 
+            session = requests.Session()
```
kmtaylor-github: In fact, is it even worth doing the typecheck? Do we just assume the caller knows what they're doing and only handle the `None` case?

pyoozie/client.py
evadb: You're too fast LOL pylint just yelled at me about this so fixing it now!

tests/pyoozie/test_client.py
evadb: I've been too spoiled by pre push hooks not letting me push silly linting errors

pyoozie/client.py
evadb: True. I thought it might be a nice way to make sure whatever is there is a session, but I guess the user will find out /really/ quickly if they aren't passing in a session

evadb: 🔥 'd it

tests/pyoozie/test_client.py
```diff
@@ -152,11 +161,23 @@ def test_construction(self, mock_test_conn, oozie_config):
         api = client.OozieClient(**oozie_config)
         assert not mock_test_conn.called
         assert api._url == 'http://localhost:11000/oozie'
+        assert api._session
+
+    @mock.patch('pyoozie.client.OozieClient._test_connection')
+    def test_contruction_custom_session(self, mock_test_conn, oozie_config):
```
kmtaylor-github: Nitpick: `s/contruction/construction/`

